PR #20, opened 2022-08-08
feat: Deduplicate config

whuang1202:
Deduplicating configuration into config file
The config file is mostly the same as the common-config given, null checks are added in to the get_producer function in order to accomodate these functions on a separate file. 

@timmc-edx when you wrote the override_settings in the test file did you intend for the name of the settings (SCHEMA_REGISTRY_URL for example) to not have EVENT_BUS_KAFKA before it? I couldn't make the tests run as intended without the EVENT_BUS_KAFKA in front of it.

### DIFF
--- a/edx_event_bus_kafka/config.py
+++ b/edx_event_bus_kafka/config.py
@@ -1,0 +1,54 @@
+"""
+Configuration loading and validation.
+"""
+
+import warnings
+from typing import Optional
+
+from confluent_kafka.schema_registry import SchemaRegistryClient
+from django.conf import settings
+
+
+def create_schema_registry_client() -> Optional[SchemaRegistryClient]:
+    """
+    Create a schema registry client from common settings.
+    """
+    url = getattr(settings, 'EVENT_BUS_KAFKA_SCHEMA_REGISTRY_URL', None)
+    if url is None:
+        warnings.warn("Cannot configure event-bus-kafka: Missing setting EVENT_BUS_KAFKA_SCHEMA_REGISTRY_URL")
+        return None
+
+    key = getattr(settings, 'EVENT_BUS_KAFKA_SCHEMA_REGISTRY_API_KEY', '')
+    secret = getattr(settings, 'EVENT_BUS_KAFKA_SCHEMA_REGISTRY_API_SECRET', '')
+
+    return SchemaRegistryClient({
+        'url': url,
+        'basic.auth.user.info': f"{key}:{secret}",
+    })
+
+
+def load_common_settings() -> Optional[dict]:
+    """
+    Load common settings, a base for either producer or consumer configuration.
+    """
+    bootstrap_servers = getattr(settings, 'EVENT_BUS_KAFKA_BOOTSTRAP_SERVERS', None)
+    if bootstrap_servers is None:
+        warnings.warn("Cannot configure event-bus-kafka: Missing setting EVENT_BUS_KAFKA_BOOTSTRAP_SERVERS")
+        return None
+
+    base_settings = {
+        'bootstrap.servers': bootstrap_servers,
+    }
+
+    key = getattr(base_settings, 'EVENT_BUS_KAFKA_API_KEY', None)
+    secret = getattr(base_settings, 'EVENT_BUS_KAFKA_API_SECRET', None)
+
+    if key and secret:
+        base_settings.update({
+            'sasl.mechanism': 'PLAIN',
+            'security.protocol': 'SASL_SSL',
+            'sasl.username': key,
+            'sasl.password': secret,
+        })
+
+    return base_settings
--- a/edx_event_bus_kafka/publishing/event_producer.py
+++ b/edx_event_bus_kafka/publishing/event_producer.py
@@ -6,16 +6,15 @@ Main function is ``send_to_event_bus``.
 
 import json
 import logging
-import warnings
 from functools import lru_cache
 from typing import Any, List, Optional
 
 from confluent_kafka import SerializingProducer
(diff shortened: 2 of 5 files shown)


edx_event_bus_kafka/publishing/event_producer.py
```diff
@@ -176,10 +155,11 @@ def inner_to_dict(event_data, ctx=None):  # pylint: disable=unused-argument
         to_dict=inner_to_dict,
     )
 
-    producer_settings.update({
-        'key.serializer': key_serializer,
-        'value.serializer': value_serializer,
-    })
+    if producer_settings is not None:
```
timmc-edx, line 158: If `producer_settings` is `None`, that means the settings weren't valid -- so we'll want to do this check and `return None` as soon as we can.

edx_event_bus_kafka/config.py
```diff
@@ -0,0 +1,59 @@
...
+
+import warnings
+import attr
+from confluent_kafka.schema_registry import SchemaRegistryClient
+from django.conf import settings
+
+
+@attr.s
+class CommonConfig:
+    settings = attr.ib(type=dict)
```
timmc-edx, line 15: This is unused. (Also it looks like you committed a `default.db` file by accident?)

edx_event_bus_kafka/config.py
```diff
@@ -0,0 +1,59 @@
...
+
+    key    = getattr(settings, 'EVENT_BUS_KAFKA_SCHEMA_REGISTRY_API_KEY', '')
+    secret = getattr(settings, 'EVENT_BUS_KAFKA_SCHEMA_REGISTRY_API_SECRET', '')
+
+    return SchemaRegistryClient({
+        'url': url,
+        'basic.auth.user.info': f"{key}:{secret}",
+    })
+
+def load_common_settings(settings) -> Optional[dict]:
```
timmc-edx, line 35: I don't know that the indirection is needed here -- can we just use the imported `settings` directly?

edx_event_bus_kafka/publishing/test_event_producer.py
```diff
@@ -65,13 +65,14 @@ def test_get_producer_for_signal(self):
             assert ep.get_producer_for_signal(signal, 'user.id') is None
             assert len(caught_warnings) == 1
             assert str(caught_warnings[0].message).startswith("Cannot configure event-bus-kafka: Missing setting ")
+            print (caught_warnings[0].message)
```
timmc-edx, line 68: There's a debugging line here that should get removed.

edx_event_bus_kafka/config.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    Load common settings, a base for either producer or consumer configuration.
+    """
+    bootstrap_servers = getattr(settings, 'EVENT_BUS_KAFKA_BOOTSTRAP_SERVERS', None)
+    if bootstrap_servers is None:
+        warnings.warn("Cannot configure event-bus-kafka: Missing setting EVENT_BUS_KAFKA_BOOTSTRAP_SERVERS")
+        return None
+
+    settings = {
```
timmc-edx, line 44: This will need a new name so that it doesn't conflict with the Django settings, and so that the next few lines are pulling from Django settings.

timmc-edx:
Yeah, in my proof-of-concept branch I had changed the setting names in the new code, but hadn't updated the tests to match. The first part of that was intentional, though. :-)

This will also want a CHANGELOG.rst entry and a major-version bump (in the root module's __init__.py).

edx_event_bus_kafka/config.py
whuang1202: when I tried to do that it said something along the lines of settings is used before declaration, possibly having to do with multiple files which makes settings not actually import when being called in the load_common_settings function

edx_event_bus_kafka/publishing/test_event_producer.py
whuang1202: will do

edx_event_bus_kafka/config.py
whuang1202: gotcha, do we want bootstrap_servers to call it from settings and then key/secret from the other settings?

whuang1202: with smth like bootstrap_settings as the name

edx_event_bus_kafka/config.py
timmc-edx: See other (newer) comment -- I had never actually run the code on that branch, so I hadn't noticed that I was both declaring  a local `settings` in the function but also using `from django.conf import settings` (which makes a global binding of `settings`). The solution is to declare a `base_settings` (or something) at which point you can remove the `settings` parameter.

edx_event_bus_kafka/publishing/event_producer.py
```diff
@@ -131,35 +130,15 @@ def get_producer_for_signal(signal: OpenEdxPublicSignal, event_key_field: str) -
...
-        producer_settings.update({
-            'sasl.mechanism': 'PLAIN',
-            'security.protocol': 'SASL_SSL',
-            'sasl.username': settings.KAFKA_API_KEY,
-            'sasl.password': settings.KAFKA_API_SECRET,
-        })
-
-    schema_registry_client = SchemaRegistryClient(schema_registry_config)
     signal_serializer = get_serializer(signal)
-
```
timmc-edx, line 162: I'm not sure why the linter is saying there's an issue on line 147, but you'll want to restore this empty line to make it happy. (Blank line before a `def`.)

timmc-edx (approved):

Issue #15, opened 2022-07-28
Change setting name to KAFKA_BOOTSTRAP_SERVER


rgraber:
Somehow the KAFKA_BOOTSTRAP_SERVER setting ended up being pluralized. This ticket is to rename it back to the original singular version.

timmc-edx:
As discussed on https://github.com/openedx/event-bus-kafka/pull/23 we should probably be using the plural. Current plan is to switch the consumer container to use it. (Setting names all got prefixed, too, so they need an update anyhow.)

PR #23, opened 2022-08-11
refactor: Consumer uses the same config loading as producer.

dianakhuang:
We also refactor the consumer to use a common config.

https://github.com/openedx/event-bus-kafka/issues/15

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/edx_event_bus_kafka/consumer/event_consumer.py
+++ b/edx_event_bus_kafka/consumer/event_consumer.py
@@ -5,7 +5,6 @@ Core consumer and event-loop code.
 import logging
 
 from confluent_kafka import DeserializingConsumer, KafkaError
-from confluent_kafka.schema_registry import SchemaRegistryClient
 from confluent_kafka.schema_registry.avro import AvroDeserializer
 from django.conf import settings
 from django.core.management.base import BaseCommand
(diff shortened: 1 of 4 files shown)


timmc-edx:
The Kafka library expects this to be a list of servers, though -- `bootstrap.servers` -- so I'm not sure why we'd want it to be singular. (Even if we only plan to put one entry in the list.)

timmc-edx:
Reference: https://kafka.apache.org/documentation/#producerconfigs_bootstrap.servers

edx_event_bus_kafka/consumer/event_consumer.py
```diff
@@ -16,6 +15,8 @@
 from openedx_events.learning.signals import SESSION_LOGIN_COMPLETED
 from openedx_events.tooling import OpenEdxPublicSignal
 
+from edx_event_bus_kafka.config import create_schema_registry_client, load_common_settings
```
timmc-edx, line 18: Oh right, we missed the actual deduplication part. :-)

timmc-edx (approved):

PR #27, opened 2022-08-22
fix: Properly load auth settings for producer/consumer

timmc-edx:
Common auth settings were ignored since 0.3.1.

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed


edx_event_bus_kafka/tests/test_config.py
```diff
@@ -0,0 +1,50 @@
...
+from unittest import TestCase
+
+from django.test.utils import override_settings
+
+from edx_event_bus_kafka import config
+
+try:
+    from confluent_kafka.schema_registry import SchemaRegistryClient
+except ImportError:  # pragma: no cover
+    SchemaRegistryClient = None  # pragma: no cover
```
robrap, line 14: See https://github.com/openedx/event-bus-kafka/blob/main/docs/decisions/0005-optional-import-of-confluent-kafka.rst. Let's use `confluent_kafka`.

edx_event_bus_kafka/tests/test_config.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class TestSchemaRegistryClient(TestCase):
+    def test_unconfigured(self):
+        assert config.create_schema_registry_client() is None
+
+    def test_configured(self):
+        with override_settings(EVENT_BUS_KAFKA_SCHEMA_REGISTRY_URL='http://localhost:12345'):
+            assert isinstance(config.create_schema_registry_client(), SchemaRegistryClient)
```
robrap, line 23: Does this test only work if the library is installed? If not, should we just pass in None or a mock, without the import that could fail?

robrap (commented):
Fly by review. I didn't look too hard at the final new tests, but could come back to this.

edx_event_bus_kafka/tests/test_config.py
timmc-edx: The tests are always run with `confluent_kafka` installed, so there's nothing that would check `if confluent_kafka`. This is just to make the module importable. I could probably even replace `SchemaRegistryClient = None` with `pass`.

edx_event_bus_kafka/tests/test_config.py
timmc-edx: It only works if the library is installed, but `test.in` does include it.

edx_event_bus_kafka/tests/test_config.py
robrap: 1. Curious where and how does this breaks if you don't have a try block. Would edx-platform complain about test code in this dependency?
2. If this is needed, maybe something like `confluent_kafka = None` accompanied by an assertion in the test that `confluent_kafka is not None` would make it more clear that tests can't run without this.

timmc-edx: What I picture is that there _might_ be some code in edx-platform that tries loading all modules... or something. I don't know. It seems safe to have the try-block there.

I don't think it's worth building up test code around it, though. Tests will fail regardless -- especially if I change it to a `pass`.

robrap: From the perspective of someone trying to understand the tests by looking at the code, I found this confusing, and am just trying to find a way to make it clear (via comment, assertion, or otherwise).

Another proposal is to add a comment like the following:
```
    # Although confluent_kafka is required for tests to pass, adding protective code
    # just in case the module is imported elsewhere.
    # See https://github.com/openedx/event-bus-kafka/blob/main/docs/decisions/0005-optional-import-of-confluent-kafka.rst
    pass
```

timmc-edx: Can you take a look at https://github.com/openedx/event-bus-kafka/pull/30? I've made it a separate PR.

robrap: Thanks. Looks good, once the other PR is merged.

robrap (approved):
Thanks. All set. Don't forget to merge the ADR update PR. :)

Issue #31, opened 2022-08-25
[Producer] Ensure that delivery callbacks are processed in timely manner (requires some discovery)


timmc-edx:
We should ensure that polling is happening at a consistent enough rate on the producer.

Currently, we call `poll(0)` every time we send an event, which means we will _eventually_ call it for almost every ack that comes back. But there are two issues we suspect with this method:

- If events are sent at a slow or variable rate, acks may take a long time to be processed as callbacks/stats/???
- The last event sent before server shutdown is guaranteed not to have its ack processed

The other recommended method for handling producer polling is to call `poll(0)` every N seconds in a separate thread. This would decouple acks/error reporting reliability from signal production rate and ensure that the last events have their acks processed (unless server shutdown happens too soon after; see https://github.com/openedx/event-bus-kafka/issues/11 for details there.)

A/C:

- Discovery:
    - Understand the implications of:
        - Failing to process some callbacks (at end of server lifecycle)
        - Delayed processing (due to infrequent events)
        - Just not calling poll at all
            - Answer: [Messages are kept in the queue](https://github.com/edenhill/librdkafka/wiki/FAQ#when-and-how-should-i-call-rd_kafka_poll) until their delivery report callback is run, so failing to call poll at all will eventually produce errors:
              > To avoid ERR__QUEUE_FULL errors, make sure to call poll regularily.
    - Pick a method, or decide we don't care about acks and delivery callbacks
- Implementation:
    - Implement the decision, along with an ADR

References:

- https://github.com/edenhill/librdkafka/wiki/FAQ#when-and-how-should-i-call-rd_kafka_poll
- https://github.com/edenhill/librdkafka/blob/master/INTRODUCTION.md#producer-message-delivery-success
- https://github.com/edenhill/librdkafka/blob/4faeb8132521da70b6bcde14423a14eb7ed5c55e/src/rdkafka.h#L3051
- https://github.com/edenhill/librdkafka/blob/f092c290995ca81b3afb4015fcc3350ba02caa96/examples/producer.c#L64
  > The callback is triggered from rd_kafka_poll() and executes on the application's thread.
- https://github.com/edenhill/librdkafka/blob/7ae0fbda54e09cbf73057ddd6f1df42603585d28/INTRODUCTION.md#threads-and-callbacks lists all the callbacks that polling triggers

timmc-edx:
[...]

PR #39, opened 2022-09-01
feat: Poll producer on a regular cadence to avoid delayed callbacks

timmc-edx:
Addresses https://github.com/openedx/event-bus-kafka/issues/31

Also an unrelated commit to fix some minor test issues I encountered.

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/edx_event_bus_kafka/internal/producer.py
+++ b/edx_event_bus_kafka/internal/producer.py
@@ -6,9 +6,13 @@ Main function is ``get_producer()``.
 
 import json
 import logging
+import threading
+import time
+import weakref
 from functools import lru_cache
 from typing import Any, List, Optional
 
+from django.conf import settings
 from django.dispatch import receiver
 from django.test.signals import setting_changed
 from openedx_events.event_bus.avro.serializer import AvroSignalSerializer
@@ -167,6 +171,13 @@ class EventProducerKafka():
     def __init__(self, producer):
         self.producer = producer
 
+        threading.Thread(
+            target=poll_indefinitely,
+            name="kafka-producer-poll",
+            args=(weakref.ref(self),),  # allow GC but also thread auto-stop (important for tests!)
+            daemon=True,  # don't block shutdown
+        ).start()
+
     def send(
             self, *, signal: OpenEdxPublicSignal, topic: str, event_key_field: str, event_data: dict,
     ) -> None:
@@ -192,13 +203,11 @@ class EventProducerKafka():
         )
 
         # Opportunistically ensure any pending callbacks from recent event-sends are triggered.
-        #
-        # This assumes events come regularly, or that we're not concerned about
-        # high latency between delivery and callback. If those assumptions are
-        # false, we should switch to calling poll(1.0) or similar in a loop on
-        # a separate thread. Or do both.
-        #
-        # Issue: https://github.com/openedx/event-bus-kafka/issues/31
+        # This ensures that we're polling at least as often as we're producing, which is a
+        # reasonable balance. However, if events are infrequent, it doesn't ensure that
+        # callbacks happen in a timely fashion, and the last event emitted before shutdown
+        # would never get a delivery callback. That's why there's also a thread calling
+        # poll(0) on a regular interval (see `poll_indefinitely`).
         self.producer.poll(0)
 
     def prepare_for_shutdown(self):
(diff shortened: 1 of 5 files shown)


edx_event_bus_kafka/internal/producer.py
```diff
@@ -210,6 +220,37 @@ def prepare_for_shutdown(self):
...
+    This ensures that callbacks are triggered in a timely fashion, rather than waiting
+    for the poll() call that we make before or after each produce() call. This may be
+    important if events are produced infrequently, and it allows the last event the
+    server emits before shutdown to have its callback run (if it happens soon enough.)
+    """
+    # The reason we hold a weakref to the whole EventProducerKafka and
+    # not directly to the Producer itself is that you just can't make
+    # a weakref to the latter (perhaps because it's a C object.)
+    while True:
+        time.sleep(1.0)
```
timmc-edx, line 238: TODO: Make configurable (setting)

edx_event_bus_kafka/internal/producer.py
```diff
@@ -210,6 +219,46 @@ def prepare_for_shutdown(self):
...
+
+        # Temporarily hold a strong ref to the producer API singleton
+        api_object = api_weakref()
+        if api_object is None:
+            return
+
+        try:
+            api_object.producer.poll(0)
+        except BaseException:
+            pass
```
rgraber, line 256: Should we log the exception here?

edx_event_bus_kafka/internal/tests/test_producer.py
```diff
@@ -137,3 +139,39 @@ def test_send_to_event_bus(self, mock_get_serializers):
...
+                    raise Exception("Exercise error handler on first iteration")
+
+            mock_producer = Mock(**{'poll.side_effect': increment_call_count})
+
+            producer_api = ep.EventProducerKafka(mock_producer)  # Created, starts polling
+
+            # Allow a little time to pass and check that the mock poll has been called
+            time.sleep(1.0)
+            assert call_count > 0
+            print(producer_api)  # Use the value here to ensure it isn't GC'd early
```
rgraber, line 162: I'm surprised pylint didn't complain about the 'print' statement. Is that just because we haven't 100% set it up?

edx_event_bus_kafka/internal/tests/test_producer.py
```diff
@@ -137,3 +139,39 @@ def test_send_to_event_bus(self, mock_get_serializers):
             on_delivery=ep.on_event_deliver,
             headers={'ce_type': 'org.openedx.learning.auth.session.login.completed.v1'},
         )
+
+    def test_polling_loop(self):
```
rgraber, line 143: Is there a way to test errors and happy path separately? Having them both here at once makes it more difficult to understand the test, and might get us in trouble if we want to start having different behavior on error as part of sad-path work.

rgraber (commented):
A few questions

edx_event_bus_kafka/internal/tests/test_producer.py
timmc-edx: Yeah, I'm a little surprised. But maybe we have that linter rule disabled/not enabled? Pylint is definitely running on this file, though.

edx_event_bus_kafka/internal/tests/test_producer.py
timmc-edx: Hmm... yeah, I think I can do that.

edx_event_bus_kafka/internal/producer.py
timmc-edx: I'm a little concerned about logging so frequently if something goes wrong. But maybe it's not _that_ frequent?

rgraber: Hm. Maybe a debug log so we can see it if we make an effort but it won't log by default on most of our production servers?

edx_event_bus_kafka/internal/tests/test_producer.py
```diff
@@ -137,3 +139,58 @@ def test_send_to_event_bus(self, mock_get_serializers):
...
+
+        # Allow garbage collection of these objects, then ask for it to happen.
+        producer_api = None
+        mock_producer = None
+        gc.collect()
+
+        # Allow things to settle down post-GC (and let thread
+        # notice the object is gone), then save off the new call
+        # count.
+        time.sleep(0.2)
```
rgraber, line 170: A comment here about how you determined 0.2 would be useful (even if it was just heuristic).

rgraber (approved):
Still think we should DEBUG log errors but I don't think it's a blocking concern. Also one nitpicky comment request

timmc-edx:
Bleagh, I accidentally squash-merged with an empty commit body. Oh well, at least the PR number is in the commit title.

PR #40, opened 2022-09-07
Fix github url strings (org edx -> openedx)

sarina:
## This PR was autogenerated

This pr replaces the old GitHub organization, github.com/edx, with the new GitHub organization, github.com/openedx.

Tagging @openedx/tcril-engineering for review, but others are welcome to provide review.

Ref: https://github.com/openedx/tcril-engineering/issues/42


pylintrc
```diff
@@ -2,7 +2,7 @@
 # ** DO NOT EDIT THIS FILE **
 # ***************************
 #
-# This file was generated by edx-lint: https://github.com/edx/edx-lint
+# This file was generated by edx-lint: https://github.com/openedx/edx-lint
```
timmc-edx, line 5: I think this file needs to be regenerated rather than edited directly.

sarina: @timmc-edx yep! A few people called this out. Could you re-review now that I've run `edx_lint`?

pylintrc_backup
```diff
@@ -2,7 +2,7 @@
 # ** DO NOT EDIT THIS FILE **
 # ***************************
 #
-# This file was generated by edx-lint: https://github.com/edx/edx-lint
```
timmc-edx, line 5: Hmm, we should just delete this file, it shouldn't be here. I can take care of that in a separate PR, though.

timmc-edx (approved):

pylintrc_backup
sarina: Hmm. In all my test repos this file never showed up, but it did in a number of other repos. I wonder if it's a `.gitignore` thing. I can remove it.

sarina: Ah, got it. Thanks!

PR #41, opened 2022-09-07
docs: Add ADR for polling loop

timmc-edx:
**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/docs/decisions/0007-producer-polling.rst
+++ b/docs/decisions/0007-producer-polling.rst
@@ -1,0 +1,30 @@
+7. Polling the producer
+#######################
+
+Status
+******
+
+Accepted
+
+Context
+*******
+
+Polling the producer is required in order to trigger delivery callbacks that are configured by the application. These callbacks are not triggered until the event is sent, ack'd, and finally polled. Even if callbacks are not configured, it may be necessary to call ``poll`` in order to burn down the delivery report queue.
+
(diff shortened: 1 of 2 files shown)


docs/decisions/0007-producer-polling.rst
```diff
@@ -0,0 +1,30 @@
...
+
+The `recommendation`_ in the librdkafka docs is to call ``poll(0)`` before or after any event production, which is a parasitic approach ensuring that we're triggering callbacks at least as often as we're producing events. The events getting callbacks will _not_ be the ones just produced, since they won't have been acknowledged yet. However, if events are infrequent, it doesn't ensure that callbacks happen in a timely fashion, and the last event emitted before shutdown would never get a delivery callback.
+
+The documentation offers a second strategy of just running a polling loop in a separate thread. This could be in addition to or instead of the parasitic polling (although the parasitic approach specifically has the advantage of automatically smoothing out polling without having to fine-tune the looping interval, so it's worth keeping in any case.) We don't have any important delivery callbacks configured at the moment, and it's not clear whether we ever will, or whether they'll be delay-sensitive—so it wasn't clear whether we should bother setting up a polling loop.
```
robrap, line 16: It feels like the following is part of the decision, and not necessarily the context:
```
(although the parasitic approach specifically has the advantage of automatically smoothing out polling without having to fine-tune the looping interval, so it's worth keeping in any case.)
```

timmc-edx: Ah yeah, good call. It was straining to escape those parentheses, too. :-)

robrap (approved):

PR #43, opened 2022-09-08
temp: sad path poc

rgraber:
This is not for actual merging, just for review of the concept of using error topics and standardized logs on the producer side. 

The idea is to first try sending events to an error topic. This is for cases where there may be a serialization problem or a missing primary key field or something. If we cannot send to the error topic either, we log in a standard format. The error topic has a very open schema so we should be able to send events that are malformed in multiple ways to the same error topic.

### DIFF
--- a/edx_event_bus_kafka/internal/producer.py
+++ b/edx_event_bus_kafka/internal/producer.py
@@ -19,6 +19,7 @@ from openedx_events.event_bus.avro.serializer import AvroSignalSerializer
 from openedx_events.tooling import OpenEdxPublicSignal
 
 from .config import get_schema_registry_client, load_common_settings
+from .exceptions import BadConfigurationException, MissingKeyException
 
 logger = logging.getLogger(__name__)
 
@@ -27,7 +28,7 @@ try:
     import confluent_kafka
     from confluent_kafka import Producer
     from confluent_kafka.schema_registry.avro import AvroSerializer
-    from confluent_kafka.serialization import MessageField, SerializationContext
+    from confluent_kafka.serialization import MessageField, SerializationContext, StringSerializer
 except ImportError:  # pragma: no cover
     confluent_kafka = None
 
@@ -53,14 +54,14 @@ def extract_event_key(event_data: dict, event_key_field: str) -> Any:
     for field_name in field_path:
         if isinstance(current_data, dict):
             if field_name not in current_data:
-                raise Exception(
+                raise MissingKeyException(
                     f"Could not extract key from event; lookup in {event_key_field} "
                     f"failed at {field_name!r} in dictionary"
                 )
             current_data = current_data[field_name]
         else:
             if not hasattr(current_data, field_name):
-                raise Exception(
+                raise MissingKeyException(
                     f"Could not extract key from event; lookup in {event_key_field} "
                     f"failed at {field_name!r} in object"
                 )
@@ -134,7 +135,7 @@ def get_serializers(signal: OpenEdxPublicSignal, event_key_field: str):
     """
     client = get_schema_registry_client()
     if client is None:
-        raise Exception('Cannot create Kafka serializers -- missing library or settings')
+        raise BadConfigurationException('Cannot create Kafka serializers -- missing library or settings')
 
     signal_serializer = AvroSignalSerializer(signal)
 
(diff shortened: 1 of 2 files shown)


edx_event_bus_kafka/internal/producer.py
```diff
@@ -210,6 +220,16 @@ def send(
         # poll(0) on a regular interval (see `poll_indefinitely`).
         self.producer.poll(0)
 
+    def _send_to_error_topic(self, *, topic: str, event_data: dict, event_key = None):
+        try:
+            event_key_as_str = str(event_key if event_key else "Missing Key")
+            event_data_as_str = json.dumps(event_data, allow_nan=True, skipkeys=True, default=lambda x: str(x))
+            self.producer.produce(f"{topic}-error", key=event_key_as_str, value=event_data_as_str)
```
timmc-edx, line 227: I think we would need to encode to bytes, here. (We don't actually have to worry about schemas for the error topic, right? This wouldn't be an SerializingProducer and wouldn't know about Avro.)

timmc-edx: It's already a regular (non-serializing) `Producer`, though -- I completed the work to separate out the serialization step. So all we need to do is convert the strings to bytes.

edx_event_bus_kafka/internal/producer.py
```diff
@@ -191,16 +194,23 @@ def send(
...
+
+            self.producer.produce(
+                topic, key=key_bytes, value=value_bytes, headers=headers, on_delivery=on_event_deliver,
+            )
+        except BadConfigurationException as bce:
+            logger.error(LOG_FORMAT_STRING.format(str(event_data), bce))
+        except MissingKeyException as mke:
+            self._send_to_error_topic(topic=topic, event_data=event_data)
+        except:
+            self._send_to_error_topic(topic=topic, event_data=event_data, event_key=extract_event_key(event_data, event_key_field))
```
robrap, line 213: Questions:
1. Will this be returning https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#kafkaerror? Might we need to investigate details of the error?
2. Do we know what types of exceptions could happen here vs in `on_event_deliver`? Does it matter?
3. Should we send exception information like in https://docs.confluent.io/cloud/current/connectors/dead-letter-queue.html#view-the-dlq error headers?
4. Should we rename to `_send_to_error_topic_or_log`? I was confused at first how we were going to get to Kafka if we just had problems getting to Kafka. I _think_ the plan here is just to try, and if it doesn't work, then we log?

edx_event_bus_kafka/internal/producer.py
rgraber: Hm, it's interesting that it worked without going down to bytes. It's still probably a better idea, but now we know we can send other things.

timmc-edx: Ah, here we go -- looks like it accepts either string or bytes: https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Producer.produce

rgraber: Do you think there's a value to one over the other? I can't see a particular advantage but at this point you know more about the guts of kafka than I do

timmc-edx: Probably doesn't matter -- some code in there is automatically encoding strings to bytes. Doing the encoding ourselves would be more explicit and so I'd favor it slightly, but I don't think it matters much either way.

edx_event_bus_kafka/internal/producer.py
rgraber: For 1/2, so far the only remotely helpful information I can find is https://github.com/confluentinc/confluent-kafka-python/blob/master/tests/test_KafkaError.py , but it's not actually as helpful as I would like it to be. It's clear some things raise and some things just pass along the error but I'm having trouble distinguishing them.
3. yes, probably a good idea
4. i put very little thought into method names, i figured they would be cleaned up in the real thing.

rgraber: Actually these tests are slightly more useful, but only slightly https://github.com/confluentinc/confluent-kafka-python/blob/master/tests/test_Producer.py

robrap:
@rgraber: Should this be closed?

PR #44, opened 2022-09-08
feat!: add KafkaEventConsumer to public API

rgraber:
Two changes, though I can be convinced to split them out.
First makes KafkaEventConsumer public
Second renames EventProducerKafka to KafkaEventProducer to be consistent.

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [ ] Commits are squashed

### DIFF
--- a/edx_event_bus_kafka/__init__.py
+++ b/edx_event_bus_kafka/__init__.py
@@ -9,4 +9,4 @@ See ADR ``docs/decisions/0006-public-api-and-app-organization.rst`` for the reas
 from edx_event_bus_kafka.internal.consumer import KafkaEventConsumer
 from edx_event_bus_kafka.internal.producer import EventProducerKafka, get_producer
 
-__version__ = '0.6.3'
+__version__ = '0.7.0'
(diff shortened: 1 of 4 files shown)


edx_event_bus_kafka/__init__.py
```diff
@@ -6,6 +6,7 @@
 See ADR ``docs/decisions/0006-public-api-and-app-organization.rst`` for the reasoning.
 """
 
+from edx_event_bus_kafka.internal.consumer import KafkaEventConsumer
```
timmc-edx, line 9: I know this is a draft PR, but I'm just now noticing that we have `EventProducerKafka` vs. `KafkaEventConsumer` -- probably want to get those names aligned.

rgraber: I like KafkaEventConsumer better. I'll change EventProducerKafka and up the version. Now that edx-event-bus-kafka is in normal dependencies it shouldn't try to bring in the upgrade before we're ready.

dianakhuang (approved):

PR #50, opened 2022-10-03
fix: send schema to fully-qualified topic when using topic prefixes

rgraber:
The event schema is sent to the topic passed to the SerializationContext used in the serialization call. We were using the un-prefixed topic, which meant that events were getting sent to topics with no schemas, making them unreadable in the Confluent UI. This fixes this issue by passing the fully-qualified topic to the SerializationContext objects.

**Merge checklist:**
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

### DIFF
--- a/edx_event_bus_kafka/internal/producer.py
+++ b/edx_event_bus_kafka/internal/producer.py
@@ -191,14 +191,14 @@ class KafkaEventProducer():
               string naming the dictionary keys to descend)
             event_data: The event data (kwargs) sent to the signal
         """
+        full_topic = get_full_topic(topic)
+
         event_key = extract_event_key(event_data, event_key_field)
         headers = {EVENT_TYPE_HEADER_KEY: signal.event_type}
 
         key_serializer, value_serializer = get_serializers(signal, event_key_field)
-        key_bytes = key_serializer(event_key, SerializationContext(topic, MessageField.KEY, headers))
-        value_bytes = value_serializer(event_data, SerializationContext(topic, MessageField.VALUE, headers))
-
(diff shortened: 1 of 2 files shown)


edx_event_bus_kafka/internal/producer.py
```diff
@@ -191,14 +191,14 @@ def send(
...
+
         event_key = extract_event_key(event_data, event_key_field)
         headers = {EVENT_TYPE_HEADER_KEY: signal.event_type}
 
         key_serializer, value_serializer = get_serializers(signal, event_key_field)
-        key_bytes = key_serializer(event_key, SerializationContext(topic, MessageField.KEY, headers))
-        value_bytes = value_serializer(event_data, SerializationContext(topic, MessageField.VALUE, headers))
-
-        full_topic = get_full_topic(topic)
+        key_bytes = key_serializer(event_key, SerializationContext(full_topic, MessageField.KEY, headers))
```
timmc-edx, line 200: Hah, whoops! I even remember thinking "man, I don't like having `topic` mean two different things in our code" so I'm surprised I didn't switch these over too.

timmc-edx: Speaking of which, I wonder if we should clarify the docstring to note that the topic parameter is the "base topic", without an environment prefix.

rgraber: that makes sense, will update

rgraber:
Commit lint is failing because I forgot to add a '!' after fixup on one of my commits, but it shouldn't matter once I squash.

PR #57, opened 2022-10-07
feat: Add monitoring to consumption tasks.

dianakhuang:
In order to have New Relic monitoring of event handling/processing, we are adding a decorator to the handling of single events.

https://github.com/edx/edx-arch-experiments/issues/70

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/edx_event_bus_kafka/internal/consumer.py
+++ b/edx_event_bus_kafka/internal/consumer.py
@@ -7,6 +7,7 @@ import logging
 from django.conf import settings
 from django.core.management.base import BaseCommand
 from django.dispatch import receiver
+from edx_django_utils.monitoring import background_task
 from edx_toggles.toggles import SettingToggle
 from openedx_events.event_bus.avro.deserializer import AvroSignalDeserializer
 from openedx_events.learning.data import UserData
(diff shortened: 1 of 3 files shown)


timmc-edx (approved):

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -114,6 +115,7 @@ def consume_indefinitely(self):
             self.consumer.close()
             logger.info("Committing final offsets")
 
+    @background_task()
```
robrap, line 118: @dianakhuang: I know this merged, but some things to consider:

1. Should the `poll` call be included in the task? Do we need a method that wraps all of the following?:
```
    msg = self.consumer.poll(timeout=CONSUMER_POLL_TIMEOUT)
    if msg is not None:
        self.process_single_message(msg)
```
2. Will all tasks have the same name? Do we need to use a context manager to set the task name, possibly using `msg.topic()` or something?
3. How can we convey when there was an error to New Relic? Do we have enough info from `msg.error()` to use `notice_error`? See https://docs.newrelic.com/docs/apm/agents/python-agent/python-agent-api/noticeerror-python-agent-api/? Note: edx-django-utils currently exposes `record_exception`, which has been deprecated in favor of notice_error. We may want to mark is as deprecated and add the new method.

PR #58, opened 2022-10-11
feat: Change `EVENT_BUS_KAFKA_CONSUMERS_ENABLED` default to True

timmc-edx:
This was originally intended as a rollout switch, but running the consumer already requires invoking a management command. Defaulting to True makes this an emergency kill-switch, which is more plausibly useful.

Remove ticket link that is now outdated.

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/edx_event_bus_kafka/internal/consumer.py
+++ b/edx_event_bus_kafka/internal/consumer.py
@@ -27,12 +27,12 @@ except ImportError:  # pragma: no cover
 
 # .. toggle_name: EVENT_BUS_KAFKA_CONSUMERS_ENABLED
 # .. toggle_implementation: SettingToggle
-# .. toggle_default: False
-# .. toggle_description: Enables the ability to listen and process events from the Kafka event bus
-# .. toggle_use_cases: opt_in
+# .. toggle_default: True
+# .. toggle_description: If set to False, consumer will exit immediately. This can be used as an emergency kill-switch
+#   to disable a consumer—as long as the management command is killed and restarted when settings change.
+# .. toggle_use_cases: opt_out
 # .. toggle_creation_date: 2022-01-31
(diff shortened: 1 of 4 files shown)


edx_event_bus_kafka/internal/consumer.py
```diff
@@ -28,12 +28,12 @@
...
-# .. toggle_default: False
-# .. toggle_description: Enables the ability to listen and process events from the Kafka event bus
-# .. toggle_use_cases: opt_in
+# .. toggle_default: True
+# .. toggle_description: If set to False, consumer will exit immediately. This can be used as an emergency kill-switch
+#   to disable a consumer—as long as the management command is killed and restarted when settings change.
+# .. toggle_use_cases: opt_out
 # .. toggle_creation_date: 2022-01-31
-# .. toggle_tickets: https://openedx.atlassian.net/browse/ARCHBOM-1992
-KAFKA_CONSUMERS_ENABLED = SettingToggle('EVENT_BUS_KAFKA_CONSUMERS_ENABLED', default=False)
```
timmc-edx, line 36: Author note: This is currently checked in the `Command`, but it's possible that `consume_indefinitely` could be called some other way. Should the check be moved from the former to the latter?

rgraber: Hm. It's a good point, but if we don't have all the proper settings setup then the command will fail even before it gets to calling `consume_indefinitely` so we do need some kind of switch at the top

dianakhuang (approved):
I think this looks good once you manage to clean up whatever conflicts might remain from my merge.

timmc-edx:
I've just noticed that this is actually one of the things that is copied into course-discovery, so we'll need to keep it set to True for now in those configs. :shrug:

PR #63, opened 2022-10-20
fix: Don't look up signal a second time in consumer

timmc-edx:
The consumer is configured with a signal to listen for, and while the `emit_signals_from_message` was validating that messages matched the signal, it was then doing a lookup on the event type anyhow. I think this was harmless but redundant, so this removes that lookup.

Also, improve consumer tests, mostly by adding assertions on logging (done as a separate commit, but will be squashed.)

**Merge checklist:**
- [ ] Version bumped
- [x] Changelog record added
- [ ] Commits are squashed

### DIFF
--- a/edx_event_bus_kafka/internal/tests/test_consumer.py
+++ b/edx_event_bus_kafka/internal/tests/test_consumer.py
@@ -10,7 +10,6 @@ from django.core.management import call_command
 from django.test import TestCase
 from django.test.utils import override_settings
 from openedx_events.learning.data import UserData, UserPersonalData
-from openedx_events.tooling import OpenEdxPublicSignal
 
 from edx_event_bus_kafka.internal.consumer import KafkaEventConsumer
 from edx_event_bus_kafka.management.commands.consume_events import Command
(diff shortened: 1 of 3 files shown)


edx_event_bus_kafka/internal/tests/test_consumer.py
```diff
@@ -130,39 +129,35 @@ def fake_poll(*args, **kwargs):
...
-            # Should just suppress exception and log
-            self.event_consumer.emit_signals_from_message(self.normal_message)
-
-        mock_lookup.assert_called_once_with(self.signal_type)
+        mock_logger.error.assert_called_once_with("Missing ce_type header on message, cannot determine signal")
         assert not self.mock_signal.send_event.called
 
-    def test_unwanted_types(self):
+    @patch('edx_event_bus_kafka.internal.consumer.logger', autospec=True)
+    def test_unwanted_types(self, mock_logger):
```
rgraber, line 150: Not a fault with this PR but while we're here, can we rename this test something a little clearer maybe
```suggestion
    def test_wrong_singal_type_in_message_header(self, mock_logger):
```

rgraber (approved):
LGTM with one request

edx_event_bus_kafka/internal/tests/test_consumer.py
timmc-edx: Yeah, absolutely.

PR #67, opened 2022-10-21
feat: start processing messages at latest offset

rgraber:
New consumers will only processes messages that have been added to the topic since the consumer group was spun up. This will prevent new consumers from trying to process a massive backlog of events that have been publishing for a while.

**Merge checklist:**
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,7 +19,7 @@ Unreleased
 
 Changed
 =======
-* Remove auto.offset.reset: earliest from consumer (new consumer groups will now consume only messages that are sent after the group was initialized)
+* Remove override of auto.offset.reset on consumer (which will default to "latest"). New consumer groups will consume only messages that are sent after the group was initialized.
 * Remove redundant lookup of signal in consumer loop (should not have any effect)
 
 [1.3.0] - 2022-10-20
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 **********
 
+[1.4.0] - 2022-10-21
+********************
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 **********
 
+[1.4.0] - 2022-10-21
+********************
+
+Changed
+=======
+* Remove auto.offset.reset: earliest from consumer (new consumer groups will now consume only messages that are sent after the group was initialized)
```
robrap, line 22: Adjusted language to match the net result, rather than the fact that you technically removed a line.
```suggestion
* Changed "auto.offset.reset" from "earliest" to "latest" (the default value). New consumer groups will now consume only messages that are sent after the group was initialized.
```

CHANGELOG.rst
```diff
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 **********
 
+[1.4.0] - 2022-10-21
+********************
+
+Changed
+=======
+* Remove override of auto.offset.reset on consumer (which will default to "latest"). New consumer groups will consume only messages that are sent after the group was initialized.
```
robrap, line 22: I think this is heavy on the removal, which ultimately doesn't matter if we aren't even saying what was removed.  Why not something like:
```suggestion
* Updated auto.offset.reset on consumer to "latest" (the default value). New consumer groups will consume only messages that are sent after the group was initialized.
```

robrap (approved):
I like my suggestion for the changelog, but also don't think this is important enough to be a blocker.

PR #69, opened 2022-10-25
docs: Ensure TODOs have ticket links as necessary

timmc-edx:
**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/edx_event_bus_kafka/internal/consumer.py
+++ b/edx_event_bus_kafka/internal/consumer.py
@@ -162,19 +162,15 @@ class KafkaEventConsumer:
 
 class ConsumeEventsCommand(BaseCommand):
     """
-    Listen for events from the event bus and log them. Only run on servers where
-    ``EVENT_BUS_KAFKA_CONSUMERS_ENABLED`` is true.
+    Management command for Kafka consumer workers in the event bus.
     """
     help = """
-    This starts a Kafka event consumer that listens to the specified topic and logs all messages it receives. Topic
-    is required.
+    Consume messages of specified signal type from a Kafka topic and send their data to that signal.
 
     example:
-        python3 manage.py cms consume_events -t user-event-debug -g user-event-consumers
+        python3 manage.py cms consume_events -t user-login -g user-activity-service \
             -s org.openedx.learning.auth.session.login.completed.v1
     """
-    # TODO: Add pointer to relevant future docs around topics and consumer groups, and potentially
-    # update example topic and group names to follow any future naming conventions.
 
     def add_arguments(self, parser):
 
@@ -182,7 +178,7 @@ class ConsumeEventsCommand(BaseCommand):
             '-t', '--topic',
             nargs=1,
             required=True,
-            help='Topic to consume'
+            help='Topic to consume (without environment prefix)'
         )
 
         parser.add_argument(
--- a/edx_event_bus_kafka/internal/consumer.py
+++ b/edx_event_bus_kafka/internal/consumer.py
@@ -70,10 +70,6 @@ class KafkaEventConsumer:
 
         schema_registry_client = get_schema_registry_client()
 
-        # TODO (EventBus):
-        # 1. Reevaluate if all consumers should listen for the earliest unprocessed offset (auto.offset.reset)
-        # 2. Ensure the signal used in the signal_deserializer is the same one sent over in the message header
-
         signal_deserializer = AvroSignalDeserializer(self.signal)
 
         def inner_from_dict(event_data_dict, ctx=None):  # pylint: disable=unused-argument
@@ -88,6 +84,7 @@ class KafkaEventConsumer:
             'value.deserializer': AvroDeserializer(schema_str=signal_deserializer.schema_string(),
                                                    schema_registry_client=schema_registry_client,
                                                    from_dict=inner_from_dict),
+            # TODO: Possibly remove or change the offset behavior https://github.com/openedx/event-bus-kafka/issues/61
         })
 
         return DeserializingConsumer(consumer_config)
@@ -106,9 +103,7 @@ class KafkaEventConsumer:
             full_topic = get_full_topic(self.topic)
             self.consumer.subscribe([full_topic])
 
-            # TODO (EventBus):
-            # 1. Is there an elegant way to exit the loop?
-            # 2. Determine if there are other errors that shouldn't kill the entire loop
+            # TODO: Handle exceptions at all https://github.com/openedx/event-bus-kafka/issues/62
             while True:
                 msg = self.consumer.poll(timeout=CONSUMER_POLL_TIMEOUT)
                 if msg is not None:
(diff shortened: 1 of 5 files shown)


edx_event_bus_kafka/internal/consumer.py
```diff
@@ -88,6 +84,7 @@ def inner_from_dict(event_data_dict, ctx=None):  # pylint: disable=unused-argume
             'value.deserializer': AvroDeserializer(schema_str=signal_deserializer.schema_string(),
                                                    schema_registry_client=schema_registry_client,
                                                    from_dict=inner_from_dict),
+            # TODO: Possibly remove or change the offset behavior https://github.com/openedx/event-bus-kafka/issues/61
```
robrap, line 87: We already switched to latest. I don't think we need a todo here. I added a comment to the issue to get it closed.

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -106,9 +103,7 @@ def consume_indefinitely(self):
             full_topic = get_full_topic(self.topic)
             self.consumer.subscribe([full_topic])
 
-            # TODO (EventBus):
-            # 1. Is there an elegant way to exit the loop?
-            # 2. Determine if there are other errors that shouldn't kill the entire loop
+            # TODO: Handle exceptions at all https://github.com/openedx/event-bus-kafka/issues/62
```
robrap, line 106: What does this comment mean? Is it meant to be a question (see possible suggestion)? Other?
```
TODO: Handle exceptions at all? See https://...
```

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -141,8 +137,6 @@ def emit_signals_from_message(self, msg):
         """
         headers = msg.headers() or []  # treat None as []
 
-        # TODO (EventBus): iterate on error handling for missing or multiple event_type headers
-        #  (headers() is actually a list of (key, value) tuples rather than a dictionary)
```
robrap, line 145: Not sure if these need to be code comments, or notes in the consumer error ticket, but:
1. Do we have the offset or other information we can add to error messages?
2. For multiple event type headers, it would be useful to add them to the error message.
3. Maybe we should "Signal not sent.", or something like that to the message?
4. Separate question: where are we committing the offset?

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -156,8 +150,8 @@ def emit_signals_from_message(self, msg):
         # CloudEvents specifies using UTF-8 for header values, so let's be explicit.
         event_type_str = event_type.decode("utf-8")
 
-        # If we get a message with the wrong signal encoding, we do not want to send it along.
-        # TODO (EventBus): Handle this particular sad path more gracefully.
+        # TODO: Accept multiple event types: https://github.com/openedx/openedx-events/issues/78
```
robrap, line 153: It may be useful to have that issue point to this code, but I don't think we need a TODO for this. We'll do this if and when we determine we need it.

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -156,8 +150,8 @@ def emit_signals_from_message(self, msg):
...
+        # TODO: Maybe raise error here? Or at least set a metric or custom attribute.
```
robrap, line 154: Not sure how to handle this in terms of comments or notes in some task, but:
- Should all of these error logs + returns just raise an error instead? Then we could log and send the error to New Relic.

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -180,11 +174,9 @@ class ConsumeEventsCommand(BaseCommand):
     example:
         python3 manage.py cms consume_events -t user-event-debug -g user-event-consumers
             -s org.openedx.learning.auth.session.login.completed.v1
-
-    # TODO (EventBus): Add pointer to relevant future docs around topics and consumer groups, and potentially
-    update example topic and group names to follow any future naming conventions.
-
     """
+    # TODO: Add pointer to relevant future docs around topics and consumer groups, and potentially
+    # update example topic and group names to follow any future naming conventions.
```
robrap, line 179: Should we just replace this with a link to https://openedx.atlassian.net/wiki/spaces/AC/pages/3508699151/How+to+start+using+the+Event+Bus for a starting place for documentation?

edx_event_bus_kafka/management/commands/produce_event.py
```diff
@@ -29,9 +29,8 @@ class Command(BaseCommand):
                     "id": 123,
                     "is_active": true,
                     "pii": {"username": "foobob", "email": "bob@foo.example", "name": "Bob Foo"}}}'
-
-    # TODO (EventBus): Potentially update example topic and group names to follow any future naming conventions.
     """
+    # TODO: Potentially update example topic and group names to follow any future naming conventions.
```
robrap, line 33: Can we just get our current topic and group name and determine if the example should be updated or not, and close this?

robrap (commented):
Thanks.

edx_event_bus_kafka/internal/consumer.py
timmc-edx: Oops, thanks, missed that in the rebase.

edx_event_bus_kafka/internal/consumer.py
timmc-edx: What I meant was "we're not handling exceptions at all (but we should)" -- too terse, though. Will reword!

edx_event_bus_kafka/internal/consumer.py
timmc-edx: Yeah, can delete.

edx_event_bus_kafka/internal/consumer.py
timmc-edx: Probably. I can fold that into the error-handling ticket(s).

timmc-edx: I've listed that in https://github.com/edx/edx-arch-experiments/issues/55 and will break it out later.

edx_event_bus_kafka/internal/consumer.py
timmc-edx: I'm not sure it's worth linking any particular place until we have something written about naming conventions.

edx_event_bus_kafka/management/commands/produce_event.py
timmc-edx: Do you mean that we should use a real in-use topic and group name here?

...actually, I just realized that there's an error here and in the first line of the docstring -- mentions of consumers and groups are both incorrect.

edx_event_bus_kafka/internal/consumer.py
timmc-edx: 1–3, added to https://github.com/edx/edx-arch-experiments/issues/55

4: We aren't committing the offset manually, and it's being done on a timer, which isn't great. https://github.com/edx/edx-arch-experiments/issues/55 includes a note on that.

edx_event_bus_kafka/internal/consumer.py
robrap: Then maybe best to drop the todo as well. We could add a todo to the doc for address naming conventions if we think this is important, or we can just leave it until someone asks.

edx_event_bus_kafka/management/commands/produce_event.py
robrap: I think what you did here works fine.

robrap (approved):

timmc-edx:
With the management command help-string fixes, this becomes more of a `fix:` PR, so I'm adding a changelog entry and version bump.

Issue #45, opened 2022-09-12
[Producer] Log failed events


rgraber:
A/C

- [ ] A developer can find all failed events with a single grep command
- [ ] Any event that fails to publish to the designated topic is logged in a standard format
- [ ] Any event that failed to publish can be reconstructed from the log

timmc-edx:
From discussion: For validation we'll try breaking the producer so we can check the logs and telemetry.

robrap:
@rgraber: Are you able to close out this ticket? If we are just waiting on New Relic issues, let's create a new `[observability]` related task in the Production Ready milestone. You can assign it to yourself while you are actively working on this. Thanks.

rgraber:
By "task" did you mean issue? Or should I add to the existing observability issue?

PR #71, opened 2022-10-26
feat: log full event data on all errors in producing

rgraber:
Include all event data when logging errors in producing so if necessary dropped events can be replayed from the logs.

Issue: https://github.com/openedx/event-bus-kafka/issues/45

**Merge checklist:**
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,15 @@ Change Log
 Unreleased
 **********
 
+[1.5.0] - 2022-10-26
+********************
+
+Changed
+=======
+* Remove redundant lookup of signal in consumer loop (should not have any effect)
+* Explicitly encode message header values as UTF-8 (no change in behavior)
+* Log full event data on all producer errors
+
 [1.4.3] - 2022-10-31
 ********************
 
@@ -41,6 +50,7 @@ Fixed
 
 Changed
 =======
+
 * Remove override of auto.offset.reset on consumer (which will default to "latest"). New consumer groups will consume only messages that are sent after the group was initialized.
 * Remove redundant lookup of signal in consumer loop (should not have any effect)
 * Explicitly encode message header values as UTF-8 (no change in behavior)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,8 +19,6 @@ Unreleased
 
 Changed
 =======
-* Remove redundant lookup of signal in consumer loop (should not have any effect)
-* Explicitly encode message header values as UTF-8 (no change in behavior)
 * Log full event data on all producer errors
 
 [1.4.3] - 2022-10-31
@@ -50,7 +48,6 @@ Fixed
 
 Changed
 =======
-
 * Remove override of auto.offset.reset on consumer (which will default to "latest"). New consumer groups will consume only messages that are sent after the group was initialized.
 * Remove redundant lookup of signal in consumer loop (should not have any effect)
 * Explicitly encode message header values as UTF-8 (no change in behavior)
--- a/edx_event_bus_kafka/internal/producer.py
+++ b/edx_event_bus_kafka/internal/producer.py
@@ -12,9 +12,11 @@ import weakref
 from functools import lru_cache
 from typing import Any, List, Optional
 
+import attr
 from django.conf import settings
 from django.dispatch import receiver
 from django.test.signals import setting_changed
+from edx_django_utils.monitoring import record_exception
 from openedx_events.event_bus.avro.serializer import AvroSignalSerializer
 from openedx_events.tooling import OpenEdxPublicSignal
 
@@ -34,6 +36,23 @@ except ImportError:  # pragma: no cover
 # CloudEvent standard name for the event type header, see
 # https://github.com/cloudevents/spec/blob/v1.0.1/kafka-protocol-binding.md#325-example
 EVENT_TYPE_HEADER_KEY = "ce_type"
+
+
+def record_producing_error(error, context):
+    """
+    Record an error in producing an event to both the monitoring system and the regular logs
+
+    Arguments:
+        error: The exception or error raised during producing
+        context: An instance of ProducingContext containing additional information about the message
+    """
+    try:
+        # record_exception() is a wrapper around a New Relic method that can only be called within an except block,
+        # so first re-raise the error
+        raise Exception(error)
+    except BaseException:
+        record_exception()
+        logger.exception(f"Error producing event to event bus. {error=!s} {context!r}")
 
 
 def extract_event_key(event_data: dict, event_key_field: str) -> Any:
@@ -157,6 +176,41 @@ def get_serializers(signal: OpenEdxPublicSignal, event_key_field: str):
     return key_serializer, value_serializer
 
 
+@attr.s(kw_only=True, repr=False)
+class ProducingContext:
+    """
+    Wrapper class to allow us to link a call to produce() with the on_event_deliver callback
+    """
+    full_topic = attr.ib(type=str, default=None)
+    event_type = attr.ib(type=str, default=None)
+    event_key = attr.ib(type=str, default=None)
+    signal = attr.ib(type=OpenEdxPublicSignal, default=None)
+    initial_topic = attr.ib(type=str, default=None)
+    event_key_field = attr.ib(type=str, default=None)
+    event_data = attr.ib(type=dict, default=None)
+
+    def __repr__(self):
+        """Create a logging-friendly string"""
+        return " ".join([f"{key}={value!r}" for key, value in attr.asdict(self).items()])
+
+    def on_event_deliver(self, err, evt):
+        """
+        Simple callback method for debugging event production
+
+        If there is any error, log all the known information about the calling context so the event can be recreated
+        and/or resent later
+
+        Arguments:
+            err: Error if event production failed
+            evt: Event that was delivered (or failed to be delivered)
+        """
+        if err is not None:
+            record_producing_error(err, self)
+        else:
+            logger.info(f"Event delivered to topic {evt.topic()}; key={evt.key()}; "
+                        f"partition={evt.partition()}")
+
+
 class KafkaEventProducer():
     """
     API singleton for event production to Kafka.
@@ -191,28 +245,41 @@ class KafkaEventProducer():
               string naming the dictionary keys to descend)
             event_data: The event data (kwargs) sent to the signal
         """
-        full_topic = get_full_topic(topic)
 
-        event_key = extract_event_key(event_data, event_key_field)
-        # Dictionary (or list of key/value tuples) where keys are strings and values are binary.
-        # CloudEvents specifies using UTF-8; that should be the default, but let's make it explicit.
-        headers = {EVENT_TYPE_HEADER_KEY: signal.event_type.encode("utf-8")}
+        # keep track of the initial arguments for recreating the event in the logs if necessary later
+        context = ProducingContext(signal=signal, initial_topic=topic, event_key_field=event_key_field,
+                                   event_data=event_data)
+        try:
+            full_topic = get_full_topic(topic)
+            context.full_topic = full_topic
 
-        key_serializer, value_serializer = get_serializers(signal, event_key_field)
-        key_bytes = key_serializer(event_key, SerializationContext(full_topic, MessageField.KEY, headers))
-        value_bytes = value_serializer(event_data, SerializationContext(full_topic, MessageField.VALUE, headers))
+            event_key = extract_event_key(event_data, event_key_field)
+            context.event_key = event_key
 
-        self.producer.produce(
-            full_topic, key=key_bytes, value=value_bytes, headers=headers, on_delivery=on_event_deliver,
-        )
+            # Dictionary (or list of key/value tuples) where keys are strings and values are binary.
+            # CloudEvents specifies using UTF-8; that should be the default, but let's make it explicit.
+            headers = {EVENT_TYPE_HEADER_KEY: signal.event_type.encode("utf-8")}
+            context.event_type = signal.event_type
 
-        # Opportunistically ensure any pending callbacks from recent event-sends are triggered.
-        # This ensures that we're polling at least as often as we're producing, which is a
-        # reasonable balance. However, if events are infrequent, it doesn't ensure that
-        # callbacks happen in a timely fashion, and the last event emitted before shutdown
-        # would never get a delivery callback. That's why there's also a thread calling
-        # poll(0) on a regular interval (see `poll_indefinitely`).
-        self.producer.poll(0)
+            key_serializer, value_serializer = get_serializers(signal, event_key_field)
+            key_bytes = key_serializer(event_key, SerializationContext(full_topic, MessageField.KEY, headers))
+            value_bytes = value_serializer(event_data, SerializationContext(full_topic, MessageField.VALUE, headers))
+            self.producer.produce(
+                full_topic, key=key_bytes, value=value_bytes, headers=headers,
+                on_delivery=context.on_event_deliver,
+            )
+
+            # Opportunistically ensure any pending callbacks from recent event-sends are triggered.
+            # This ensures that we're polling at least as often as we're producing, which is a
+            # reasonable balance. However, if events are infrequent, it doesn't ensure that
+            # callbacks happen in a timely fashion, and the last event emitted before shutdown
+            # would never get a delivery callback. That's why there's also a thread calling
+            # poll(0) on a regular interval (see `poll_indefinitely`).
+            self.producer.poll(0)
+        except BaseException as e:
+            # Errors caused by the produce call should be handled by the on_delivery callback.
+            # Here we might expect serialization errors, or any errors from preparing to produce.
+            record_producing_error(e, context)
 
     def prepare_for_shutdown(self):
         """
@@ -251,14 +318,11 @@ def poll_indefinitely(api_weakref: KafkaEventProducer):
         api_object = api_weakref()
         if api_object is None:
             return
-
         try:
             api_object.producer.poll(0)
         except BaseException:
-            # If polling is failing, we'll almost certainly find out about it from the poll call
-            # we make when producing an event. The call in this loop could be excessively noisy,
-            # so just debug-log it.
-            logger.debug("Event bus producer polling loop encountered exception (continuing)", exc_info=True)
+            logger.exception("Event bus producer polling loop encountered exception (continuing)")
+            record_exception()
         finally:
             # Get rid of that strong ref again
             api_object = None
@@ -284,25 +348,6 @@ def get_producer() -> Optional[KafkaEventProducer]:
         return None
 
     return KafkaEventProducer(Producer(producer_settings))
-
-
-def on_event_deliver(err, evt):
-    """
-    Simple callback method for debugging event production
-
-    Arguments:
-        err: Error if event production failed
-        evt: Event that was delivered (or failed to be delivered)
-
-    Note: This is meant to be temporary until we implement
-      more rigorous error handling.
-    """
-    if err is not None:
-        logger.warning(f"Event delivery failed: {err!r}")
-    else:
-        # Don't log msg.value() because it may contain userids and/or emails
-        logger.info(f"Event delivered to topic {evt.topic()}; key={evt.key()}; "
-                    f"partition={evt.partition()}")
 
 
 @receiver(setting_changed)
--- a/edx_event_bus_kafka/internal/producer.py
+++ b/edx_event_bus_kafka/internal/producer.py
@@ -191,7 +191,7 @@ class ProducingContext:
 
     def __repr__(self):
         """Create a logging-friendly string"""
-        return " ".join([f"{key}={value!r}" for key, value in attr.asdict(self).items()])
+        return " ".join([f"{key}={value!r}" for key, value in attr.asdict(self, recurse=False).items()])
 
     def on_event_deliver(self, err, evt):
         """
--- a/edx_event_bus_kafka/internal/tests/test_producer.py
+++ b/edx_event_bus_kafka/internal/tests/test_producer.py
@@ -6,12 +6,13 @@ import gc
 import time
 import warnings
 from unittest import TestCase
-from unittest.mock import Mock, call, patch
+from unittest.mock import ANY, Mock, call, patch
 
 import openedx_events.learning.signals
 import pytest
 from django.test import override_settings
 from openedx_events.event_bus.avro.serializer import AvroSignalSerializer
+from openedx_events.event_bus.avro.tests.test_utilities import SubTestData0, create_simple_signal
 from openedx_events.learning.data import UserData, UserPersonalData
 
 import edx_event_bus_kafka.internal.producer as ep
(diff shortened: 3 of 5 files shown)


CHANGELOG.rst
```diff
@@ -14,14 +14,23 @@ Change Log
...
 
-[1.4.0] - 2022-10-21
+[1.5.0] - 2022-10-26
 ********************
 
 Changed
 =======
-* Remove override of auto.offset.reset on consumer (which will default to "latest"). New consumer groups will consume only messages that are sent after the group was initialized.
+
 * Remove redundant lookup of signal in consumer loop (should not have any effect)
```
timmc-edx, line 23: Looks like some lines here got moved.

edx_event_bus_kafka/internal/producer.py
```diff
@@ -191,28 +234,42 @@ def send(
...
+            # would never get a delivery callback. That's why there's also a thread calling
+            # poll(0) on a regular interval (see `poll_indefinitely`).
+            self.producer.poll(0)
+        except BaseException as e:
+            # log the full exception for debugging purposes
+            logger.exception(e)
+            # separately log the original event for replay later if necessary
+            logger.error(
+                ERROR_STRING_TEMPLATE.format(error=e, context=context)  # pylint: disable=logging-format-interpolation
+            )
```
robrap, line 272: 1. Logger.exception is an error log + adds the current exception, so it is all you need. See https://docs.python.org/3/library/logging.html#logging.Logger.exception.
2. I added a proposed comment, but feel free to reword. I just wanted to clarify what types of errors would happen where.
```suggestion
            # Errors caused by the produce call should be handled by the on_delivery callback.
            # Here we might expect serialization errors, or any errors from preparing to produce.
            logger.exception(
                ERROR_STRING_TEMPLATE.format(error=e, context=context)  # pylint: disable=logging-format-interpolation
            )
```

edx_event_bus_kafka/internal/producer.py
```diff
@@ -191,28 +234,42 @@ def send(
...
+
+            # Opportunistically ensure any pending callbacks from recent event-sends are triggered.
+            # This ensures that we're polling at least as often as we're producing, which is a
+            # reasonable balance. However, if events are infrequent, it doesn't ensure that
+            # callbacks happen in a timely fashion, and the last event emitted before shutdown
+            # would never get a delivery callback. That's why there's also a thread calling
+            # poll(0) on a regular interval (see `poll_indefinitely`).
+            self.producer.poll(0)
+        except BaseException as e:
```
robrap, line 266: I am going to make some `record_exception()` requests, that I'd like added to the ticket. Feel free to spin some or all of this to a new ticket. It's simple to discuss here because we are looking at all the right places.

In this case, we are in an exception handler, so we just need a simple call to `record_exception()`. See [example call](https://github.com/openedx/edx-enterprise/blob/0f21072f25f2a5f2fda7117e1c260fc148a657d5/enterprise/views.py#L708).

edx_event_bus_kafka/internal/producer.py
```diff
@@ -251,14 +308,10 @@ def poll_indefinitely(api_weakref: KafkaEventProducer):
...
             return
-
         try:
             api_object.producer.poll(0)
         except BaseException:
-            # If polling is failing, we'll almost certainly find out about it from the poll call
-            # we make when producing an event. The call in this loop could be excessively noisy,
-            # so just debug-log it.
-            logger.debug("Event bus producer polling loop encountered exception (continuing)", exc_info=True)
+            logger.warning("Event bus producer polling loop encountered exception (continuing)", exc_info=True)
```
robrap, line 314: 1. Why is this not considered an error?
```suggestion
            logger.exception("Event bus producer polling loop encountered exception (continuing)")
```
2. Can we add a call to `record_exception()`? See other related comment.

edx_event_bus_kafka/internal/producer.py
```diff
@@ -157,6 +167,39 @@ def inner_to_dict(event_data, ctx=None):  # pylint: disable=unused-argument
...
+
+        If there is any error, log all the known information about the calling context so the event can be recreated
+        and/or replayed later
+
+        Arguments:
+            err: Error if event production failed
+            evt: Event that was delivered (or failed to be delivered)
+        """
+        if err is not None:
+            logger.warning(
```
robrap, line 195: 1. I think this should be an error. Why just a warning?
2. I'd like to add a call to `record_exception()`. This could be a "TODO" comment and new task. That call needs to be in the context of exception handling, so we might need to both raise a chained exception and catch it. Maybe a helper method would keep this looking clean?

edx_event_bus_kafka/internal/producer.py
```diff
@@ -157,6 +167,39 @@ def inner_to_dict(event_data, ctx=None):  # pylint: disable=unused-argument
...
+                ERROR_STRING_TEMPLATE.format(error=err, context=self)  # pylint: disable=logging-format-interpolation
+            )
+        else:
+            logger.info(f"Event delivered to topic {evt.topic()}; key={evt.key()}; "
```
robrap, line 199: I imagine this would be too noisy for logs. Maybe debug? We could also wrap it in a toggle, if you think it is important for now.

edx_event_bus_kafka/internal/tests/test_producer.py
```diff
@@ -6,12 +6,13 @@
 import time
 import warnings
 from unittest import TestCase
-from unittest.mock import Mock, call, patch
+from unittest.mock import ANY, Mock, call, patch
```
robrap, line 9: **Note to self (@robrap):** Review tests when you return to this.

edx_event_bus_kafka/internal/producer.py
```diff
@@ -34,6 +34,16 @@
 # CloudEvent standard name for the event type header, see
 # https://github.com/cloudevents/spec/blob/v1.0.1/kafka-protocol-binding.md#325-example
 EVENT_TYPE_HEADER_KEY = "ce_type"
+ERROR_STRING_TEMPLATE = "Error publishing event to event bus. error={error} {context}"
```
timmc-edx, line 37: It makes me mildly nervous to have a format string divorced from the format call that has to define the error/context parameters. (A linter won't notice a mismatch. If we had complete test coverage I guess it would be fine though.) Perhaps it should be a function? It could call logger.error and logger.record_exception.

edx_event_bus_kafka/internal/producer.py
```diff
@@ -157,6 +167,39 @@ def inner_to_dict(event_data, ctx=None):  # pylint: disable=unused-argument
...
+    Wrapper class to allow us to link a call to produce() with the on_event_deliver callback
+    """
+    def __init__(self, context_data):
+        self.context_data = context_data
+
+    def __str__(self):
+        return " ".join([f"{key}={value}" for key, value in self.context_data.items()])
+
+    def set_context_data(self, key, value):
+        self.context_data[key] = value
```
timmc-edx, line 181: Rather than a dictionary, can ProducingContext just be an attrs class or dataclass with these keys as attributes? I think it might save a bunch of code. Wouldn't need as many string constants either.

edx_event_bus_kafka/internal/producer.py
```diff
@@ -191,28 +234,42 @@ def send(
               string naming the dictionary keys to descend)
             event_data: The event data (kwargs) sent to the signal
         """
-        full_topic = get_full_topic(topic)
 
-        event_key = extract_event_key(event_data, event_key_field)
-        # Dictionary (or list of key/value tuples) where keys are strings and values are binary.
-        # CloudEvents specifies using UTF-8; that should be the default, but let's make it explicit.
```
timmc-edx, line 198: Comment removal intentional?

edx_event_bus_kafka/internal/producer.py
```diff
@@ -191,28 +234,42 @@ def send(
...
+            context.set_context_data(EVENT_KEY_DICT_KEY, event_key)
+            headers = {EVENT_TYPE_HEADER_KEY: signal.event_type.encode("utf-8")}
+            context.set_context_data(EVENT_TYPE_DICT_KEY, signal.event_type)
+
+            key_serializer, value_serializer = get_serializers(signal, event_key_field)
+            key_bytes = key_serializer(event_key, SerializationContext(full_topic, MessageField.KEY, headers))
+            value_bytes = value_serializer(event_data, SerializationContext(full_topic, MessageField.VALUE, headers))
+            self.producer.produce(
+                full_topic, key=key_bytes, value=value_bytes, headers=headers,
+                on_delivery=context.on_event_deliver,
```
timmc-edx, line 256: ...whoa, I didn't realize this produced a bound method rather than an unbound function. Convenient.

docs/decisions/0008-error-logging-on-send.rst
```diff
@@ -0,0 +1,26 @@
...
+*******
+
+There are numerous strategies for error handling in Kafka, variously optimized for different goals (consistency, reliability, etc.). Some error-handling is built in to the confluent-kafka library, such as automatic retries when producing events. Other errors may occur outside the Kafka code, for example when trying to extract an event key from an event data dictionary.
+
+Decision
+********
+
+When dealing with errors in sending events to the event bus, we will prioritize ensuring that no events are irretrievably dropped. To that end, we will catch exceptions and log the full event data in the error message in a consistent, parseable way.
+
+We will defer any idea of using a retry queue. Also, note that this decision does not cover how we will handle errors on the consumer.
```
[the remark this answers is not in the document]
rgraber: I think we should make that decision now so we don't have an ambiguous ADR lying around until we figure out consuming. If we decide this is the ADR for both, we should update the title and make it provisional. Otherwise, I think we should leave it as accepted.

edx_event_bus_kafka/internal/producer.py
rgraber: I rather like having this for now. I also don't think it will be all that noisy. If it becomes so, we can add a toggle later.

docs/decisions/0008-error-logging-on-send.rst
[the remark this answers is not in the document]
rgraber: I'll make it provisional and update the A/C on the consuming issue to make updating the ADR part of it

CHANGELOG.rst
```diff
@@ -14,14 +14,23 @@ Change Log
...
 * Explicitly encode message header values as UTF-8 (no change in behavior)
```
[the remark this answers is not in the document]
rgraber: [fixed]

docs/decisions/0008-error-logging-on-send.rst
```diff
@@ -0,0 +1,26 @@
+8. Error logging on send
```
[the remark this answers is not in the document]
rgraber: [fixed]

docs/decisions/0008-error-logging-on-send.rst
```diff
@@ -0,0 +1,26 @@
...
+
+Context
+*******
+
+There are numerous strategies for error handling in Kafka, variously optimized for different goals (consistency, reliability, etc.). Some error-handling is built in to the confluent-kafka library, such as automatic retries when producing events. Other errors may occur outside the Kafka code, for example when trying to extract an event key from an event data dictionary.
+
+Decision
+********
+
+When dealing with errors in sending events to the event bus, we will prioritize ensuring that no events are irretrievably dropped. To that end, we will catch exceptions and log the full event data in the error message in a consistent, parseable way.
```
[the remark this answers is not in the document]
rgraber: [fixed]

docs/decisions/0008-error-logging-on-send.rst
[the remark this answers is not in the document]
rgraber: [fixed] though @rgraber should still update the consuming issue

edx_event_bus_kafka/internal/producer.py

edx_event_bus_kafka/internal/producer.py
rgraber: [fixed]

docs/decisions/0008-error-logging-on-send.rst
```diff
@@ -0,0 +1,26 @@
...
+
+Status
+******
+
+Accepted
+
+Context
+*******
+
+There are numerous strategies for error handling in Kafka, variously optimized for different goals (consistency, reliability, etc.). Some error-handling is built in to the confluent-kafka library, such as automatic retries when producing events. Other errors may occur outside the Kafka code, for example when trying to extract an event key from an event data dictionary.
```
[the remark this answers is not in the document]
rgraber: [fixed]

edx_event_bus_kafka/internal/producer.py
rgraber: [wontfix unless you want to argue about it]

edx_event_bus_kafka/internal/producer.py

edx_event_bus_kafka/internal/producer.py
rgraber: Nope, good catch

rgraber: [fixed]

edx_event_bus_kafka/internal/producer.py

CHANGELOG.rst
rgraber: [fixed]

edx_event_bus_kafka/internal/producer.py
```diff
@@ -34,16 +36,20 @@
...
-EVENT_TYPE_DICT_KEY = "event_type"
-EVENT_KEY_DICT_KEY = "event_key"
-SIGNAL_DICT_KEY = "signal"
-INITIAL_TOPIC_DICT_KEY = "initial_topic"
-EVENT_KEY_FIELD_DICT_KEY = "event_key_field"
-EVENT_DATA_DICT_KEY = "event_data"
+
+def record_error(error, context):
+    """
+    Record an error in producing and event to both the monitoring system and the regular logs
```
timmc-edx, line 43: nit: typo
```suggestion
    Record an error in producing an event to both the monitoring system and the regular logs
```

edx_event_bus_kafka/internal/producer.py
```diff
@@ -34,16 +36,20 @@
...
+    """
+    try:
+        raise Exception(error)
+    except BaseException:
+        record_exception()
+        error_template = "Error publishing event to event bus. error={error} {context!r}"
+        logger.exception(
+            error_template.format(error=error, context=context)  # pylint: disable=logging-format-interpolation
+        )
```
timmc-edx, line 52: I think this could be simplified to

```suggestion
        logger.exception(f"Error publishing event to event bus. {error=} {context!r}")
```

edx_event_bus_kafka/internal/producer.py
```diff
@@ -167,18 +173,22 @@ def inner_to_dict(event_data, ctx=None):  # pylint: disable=unused-argument
...
+    full_topic = attr.ib(type=str, default=None)
+    event_type = attr.ib(type=str, default=None)
+    event_key = attr.ib(type=str, default=None)
+    signal = attr.ib(type=OpenEdxPublicSignal, default=None)
+    initial_topic = attr.ib(type=str, default=None)
+    event_key_field = attr.ib(type=str, default=None)
+    event_data = attr.ib(type=dict, default=None)
+
+    def __repr__(self):
+        """Create a splunk-friendly string"""
```
timmc-edx, line 190: Splunk is 2U-specific -- perhaps this?
```suggestion
        """Create a logging-friendly string"""
```

edx_event_bus_kafka/internal/producer.py
rgraber: Not if I want to keep the pylint disable. it ends up going over the line limit

edx_event_bus_kafka/internal/producer.py
```diff
@@ -167,18 +173,22 @@ def inner_to_dict(event_data, ctx=None):  # pylint: disable=unused-argument
...
+        return " ".join([f"{key}={value}" for key, value in attr.asdict(self).items()])
```
timmc-edx, line 191: I wonder if we should be using `{value!r}` to make this easier to reconstruct.

edx_event_bus_kafka/internal/producer.py
rgraber: Oh wait I misread. I'll try that.

edx_event_bus_kafka/internal/producer.py
```diff
@@ -311,7 +316,8 @@ def poll_indefinitely(api_weakref: KafkaEventProducer):
         try:
             api_object.producer.poll(0)
         except BaseException:
-            logger.warning("Event bus producer polling loop encountered exception (continuing)", exc_info=True)
+            logger.exception("Event bus producer polling loop encountered exception (continuing)", exc_info=True)
```
timmc-edx, line 319: I think we don't need exc_info when using the exception method
```suggestion
            logger.exception("Event bus producer polling loop encountered exception (continuing)")
```

edx_event_bus_kafka/internal/tests/test_producer.py
```diff
@@ -167,10 +171,8 @@ def test_full_event_data_present_in_key_extraction_error(self, mock_logger, *arg
             producer_api.send(signal=simple_signal, topic='topic', event_key_field='bad_field',
                               event_data={'test_data': SubTestData0(sub_name="name", course_id="id")})
 
-        # extract the error message that was produced and check it has all relevant information (order isn't guaranteed
-        # and doesn't actually matter)
-        (error_string,) = mock_logger.error.call_args.args
-        assert "event_data={'test_data': SubTestData0(sub_name='name', course_id='id')}" in error_string
+        (error_string,) = mock_logger.exception.call_args.args
```
timmc-edx, line 174: Nice way of asserting the tuple length and destructuring it at the same time.

edx_event_bus_kafka/internal/producer.py

docs/decisions/0008-error-logging-on-send.rst
```diff
@@ -0,0 +1,26 @@
...
+********
+
+When dealing with errors in sending events to the event bus, we will prioritize ensuring that no events are irretrievably dropped. To that end, we will catch exceptions and log the full event data in the error message in a consistent, parseable way.
+
+We will defer any idea of using a retry queue. Also, note that this decision does not cover how we will handle errors on the consumer.
+
+Consequences
+************
+
+Failing to send an event to the event bus will not fail the request.
```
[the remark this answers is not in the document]
rgraber: [fixed]

docs/decisions/0008-error-logging-on-send.rst
```diff
@@ -0,0 +1,26 @@
...
+
+Recovering failed events will require retrieving and parsing server logs. If other events have been successfully processed in the meantime, these recovered events may be processed out of order.
```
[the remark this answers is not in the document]
rgraber: [fixed]

edx_event_bus_kafka/internal/producer.py
```diff
@@ -157,6 +167,39 @@ def inner_to_dict(event_data, ctx=None):  # pylint: disable=unused-argument
...
+
+    def set_context_data(self, key, value):
+        self.context_data[key] = value
+
+    def on_event_deliver(self, err, evt):
+        """
+        Simple callback method for debugging event production
+
+        If there is any error, log all the known information about the calling context so the event can be recreated
+        and/or replayed later
```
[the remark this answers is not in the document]

edx_event_bus_kafka/internal/producer.py

edx_event_bus_kafka/internal/producer.py
```diff
@@ -36,6 +38,17 @@
 EVENT_TYPE_HEADER_KEY = "ce_type"
 
 
+def record_error(error, context):
```
robrap, line 41: To make it more clear why this isn't used for some of the other errors, maybe:
1. Rename to record_event_producing_error?
2. Please add arguments to docstring. Especially helpful to know what context is meant to be.

edx_event_bus_kafka/internal/producer.py
```diff
@@ -36,6 +38,17 @@
...
+    """
+    Record an error in producing an event to both the monitoring system and the regular logs
+    """
+    try:
+        raise Exception(error)
+    except BaseException:
+        record_exception()
+        logger.exception(f"Error publishing event to event bus. {error=!s} {context!r}")
```
robrap, line 49: We use the term "produce" everywhere, except in the error message. This is the only place we use "publish"? Can you use some form of "producing" instead?

edx_event_bus_kafka/internal/producer.py
robrap: I'm fine leaving this for now. Keep in mind that this is a single event, and I'm hopeful that the Event Bus will replace most server-to-server communication in the future, so I am hoping this gets bad. We'll see.

Temporarily marking unresolved, just so you see this, not for you to take action.

robrap (commented):
Great. Thanks. Minor comments.

edx_event_bus_kafka/internal/producer.py
```diff
@@ -157,6 +170,41 @@ def inner_to_dict(event_data, ctx=None):  # pylint: disable=unused-argument
...
+    event_type = attr.ib(type=str, default=None)
+    event_key = attr.ib(type=str, default=None)
+    signal = attr.ib(type=OpenEdxPublicSignal, default=None)
+    initial_topic = attr.ib(type=str, default=None)
+    event_key_field = attr.ib(type=str, default=None)
+    event_data = attr.ib(type=dict, default=None)
+
+    def __repr__(self):
+        """Create a logging-friendly string"""
+        return " ".join([f"{key}={value!r}" for key, value in attr.asdict(self).items()])
```
timmc-edx, line 188: Could we just omit this method and drop the `repr=False`?

timmc-edx: Hmm... I tried this locally and got a test failure. In one of the unit tests, the expected substring `event_data={'test_data': {'sub_name': 'name', 'course_id': 'id'}}` became `event_data={'test_data': SubTestData0(sub_name='name', course_id='id')}` and I'm not sure why. That test-data object should be getting repr'd either way...

timmc-edx: Weird:
```
(Pdb) attr.asdict(self)['event_data']
{'test_data': {'sub_name': 'name', 'course_id': 'ABCx'}}
(Pdb) self.event_data
{'test_data': SubTestData0(sub_name='name', course_id='ABCx')}
```
OK, turns out that `asdict` recurses into values by default and converts them too! Not what I would have expected at all, and I don't think we want that? https://www.attrs.org/en/stable/api.html#attr.asdict

rgraber: I'm not convinced it matters either way. Also I like having the "=" sign in there so it's more in line with normal splunk logging conventions.

timmc-edx: Allowing attrs to define `__repr__` for us would mean having less code, but I don't feel strongly about it.

But if we keep it I think I'd like to include `recurse=False` in the call to `asdict` since as it is we're losing type information that could be important for debugging.

edx_event_bus_kafka/internal/producer.py

edx_event_bus_kafka/internal/producer.py

edx_event_bus_kafka/internal/producer.py
rgraber: I feel strongly about the = thing, but I'll add the recurse=False bit

robrap (approved):

CHANGELOG.rst
```diff
@@ -14,6 +14,15 @@ Change Log
 Unreleased
 **********
 
+[1.5.0] - 2022-10-26
+********************
+
+Changed
+=======
+* Remove redundant lookup of signal in consumer loop (should not have any effect)
+* Explicitly encode message header values as UTF-8 (no change in behavior)
```
timmc-edx, line 23: A couple of lines here were duplicated from below.

CHANGELOG.rst
```diff
@@ -41,6 +50,7 @@ Fixed
 
 Changed
 =======
+
```
timmc-edx, line 53: Nit: Stray line addition. (I don't have an opinion on the formatting, but I'm guessing it was unintentional, from a merge resolution).

timmc-edx (approved):
Looks great, thank you!

(And just a reminder to update the changelog date before merge.)

Issue #62, opened 2022-10-20
[Consumer] Catch and log exceptions in the consumer loop


timmc-edx:
Currently (v1.2.0) any exception raised in the consumer loop will kill the loop. We probably don't want errors to kill the loop, as not all deployments will necessarily be restarting the maintenance command on failure. (K8s might also let it die permanently if it crashes repeatedly.)

A/C:

- [x]  try-block at highest level of consumer loop prevents any exception from killing the loop
- [x] [Note: this only applies if https://github.com/openedx/event-bus-kafka/issues/72 is done and we are doing manual commits] Verify consumer offset commits are still sent even when an exception is caught 
- [x] When a message is consumed but cannot be processed or contains an error (ie msg.error() is not null), log the offset, full-topic, key, partition, consumer group, and any other necessary information at the error level so that later replaying would be possible.
   - possibly continue to ignore EOF_PARTITION errors, look at what that actually is
   - consider logging error properties if they may be helpful for debugging
- [x] Log the identity of any failing receivers (and the key, topic, etc.) if we get as far as sending to the signal but one or more receivers throws. (See return value of `send_event`.)
- [x] Update PII-handling ADR to mention that we're also logging PII but that we don't think it's a big concern, because logs are already handled as sensitive information and it would only happen on error.
- [x] Call edx-django-utils `record_exception` (or whatever it gets renamed to) when catching any exception (not including failing receivers)
- [x] Update ADR about this decision to catch and log all exceptions. (Also see producer ADR: https://github.com/openedx/event-bus-kafka/blob/dd96ab4f678653f3aa537d1fb446f2db10cfb094/docs/decisions/0008-baseline-error-handling-for-producer.rst) Possible information to include or consider:
  - We may use DLQs later
  - We may at some point have enough experience to handle some exceptions differently from others
  - There are a bunch of different possible kinds of errors:
    - Kafka errors
      - Poll call fails
      - Poll returns a Message but it has a Kafka error code rather than a value
    - Envelope/encoding issues
      - Header that doesn't match the expected signal type (could be either producer or consumer misconfiguration)
      - Other envelope errors (random bugs like duplicated headers or something)
      - Failure to talk to schema registry
      - Deserialization failures (schema mismatches, library version mismatches, serialization bugs in producer...)
    - Signal receiver errors
[...]

PR #76, opened 2022-11-02
feat: Put in error handling for consumer

timmc-edx:
- Wrap contents of consumer loop in a try-block and convert all exceptions into log messages and telemetry. An error will no longer kill the loop.
- Change some error cases from logging to throwing (so that our error logging and reporting is unified)
- Remove handling of msg.error (delete `process_single_message` entirely) since `DeserializingConsumer` does that for us.
- Add partition and offset to FakeMessage, add more type annotations
- Protection against fast error-looping if broker is down

Also:

- Clarify some error log messages
- Add internal utility to shut down consumer loop (needed for tests)
- Improve test coverage a little

This addresses part of https://github.com/openedx/event-bus-kafka/issues/62

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed


edx_event_bus_kafka/internal/consumer.py
```diff
@@ -85,45 +96,54 @@ def inner_from_dict(event_data_dict, ctx=None):  # pylint: disable=unused-argume
...
+                msg = None
+                try:
+                    msg = self.consumer.poll(timeout=CONSUMER_POLL_TIMEOUT)
+                    # DeserializingConsumer.poll() always returns either a valid message
+                    # or None, and raises an exception in all other cases. This means
+                    # we don't need to check msg.error() ourselves.
+                    if msg is not None:
+                        self.emit_signals_from_message(msg)
+                except BaseException as e:
+                    self.record_event_consuming_error(run_context, e, msg)
```
timmc-edx, line 141: [Author's note] I'm a little concerned that we could end up in a tight error-loop here under some conditions. May not be worth worrying about, though -- it would burn CPU, but not affect correctness.

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -85,45 +96,54 @@ def inner_from_dict(event_data_dict, ctx=None):  # pylint: disable=unused-argume
 
         return DeserializingConsumer(consumer_config)
 
+    def _shut_down(self):
```
timmc-edx, line 99: [Author's note] Currently private, but could later expose if needed.

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -85,45 +96,54 @@ def inner_from_dict(event_data_dict, ctx=None):  # pylint: disable=unused-argume
...
         try:
             full_topic = get_full_topic(self.topic)
+            run_context = {
+                'full_topic': full_topic,
+                'consumer_group': self.group_id,
+                'expected_signal': self.signal,
+            }
             self.consumer.subscribe([full_topic])
+            logger.info(
+                f"Running consumer for {run_context!r}; assigned {self.consumer.assignment()!r}"
```
timmc-edx, line 124: [Author's note] I haven't yet tested this in devstack.

docs/decisions/0009-baseline-error-handling-for-consumer.rst
```diff
@@ -0,0 +1,48 @@
...
+As we gain experience, we can add special handling for certain known cases that would be better served by a backoff, retry, or halt.
+
+Receiver exceptions are not *raised*, per se, but are returned by ``send_robust``. We will log these but not treat these as consumer errors, as these exceptions are much more likely to be related to IDA business logic than to the Event Bus itself. There may also be multiple receivers, and failures in one do not indicate that other receivers should stop receiving the event stream.
+
+All raised errors will also be sent to telemetry (New Relic) for monitoring.
+
+Consequences
+************
+
+The consumer will err on the side of low latency between IDAs, creating higher (or at least sooner) inter-IDA consistency at the possible expense of inter-object consistency within a topic (if some events are dropped and other processed). We will have to ensure that we capture all failures sufficiently durably that we can replay them, and also that receivers are capable of "healing" their data when a topic is replayed.
```
robrap, line 43:
```suggestion
The consumer will err on the side of low latency between IDAs, creating higher (or at least sooner) inter-IDA consistency at the possible expense of inter-object consistency within a topic (if some events are dropped and others processed). We will have to ensure that we capture all failures sufficiently durably that we can replay them, and also that receivers are capable of "healing" their data when a topic is replayed.
```

docs/decisions/0009-baseline-error-handling-for-consumer.rst
```diff
@@ -0,0 +1,48 @@
...
+
+Rejected Alternatives
```
robrap, line 45: Not sure whether to call this Deferred/Rejected Alternatives, but it would be great to see something around the following. I think it is especially important to use the term Dead Letter Queue, because it may be an expected piece of functionality, and it would be good to clarify explicitly, and not just implicitly.

* It is a common practice to introduce a new topic as a Dead Letter Queue (DLQ), or a place to send events that we were unable to process. For now, the log will serve as our DLQ, but a topic of messages may be simpler to re-process (although the messages will no longer be ordered). This work has been deferred as of the initial draft of this ADR.
* Additionally, sometimes a topic or topics are introduced for retrying events before the events are moved to the DLQ. Again, there are design considerations required to maintain order. This work has been deferred as of the initial draft of this ADR.

edx_event_bus_kafka/internal/tests/test_consumer.py
```diff
@@ -88,46 +102,58 @@ def setUp(self):
...
-            # for this test.
-            if poll_call_count >= 3:
+        emit_call_count = 0
+
+        def fake_emit(*args, **kwargs):
+            nonlocal emit_call_count, self
+            emit_call_count += 1
+            # Perform the following sequence of events on successive calls:
+            # 1) accept a message normally, 2) throw, 3) accept a message normally
+            # again, and 4) stop the loop so we can finish the unit test.
```
robrap, line 124: Nit: This might make more sense as a docstring?

edx_event_bus_kafka/internal/tests/test_consumer.py
```diff
@@ -88,46 +102,58 @@ def setUp(self):
...
         # Check that each of the mocked out methods got called as expected.
         mock_consumer.subscribe.assert_called_once_with(['prod-some-topic'])
-        assert mock_consumer.poll.call_args_list == [
-            call(timeout=1.0), call(timeout=1.0), call(timeout=1.0)
-        ]
-        assert mock_process.call_args_list == [
-            call(self.normal_message), call(self.normal_message)
-        ]
-        mock_consumer.close.assert_called_once_with()
+        assert mock_emit.call_args_list == [call(self.normal_message)] * 4
```
robrap, line 141: Nit: Maybe use a variable to clarify the 4. I'm not even sure if the following is accurate:
```
num_of_expected_calls = 4
```
Did you notice that `num` is an abbreviation? So much for consistency. ;)

robrap (commented):
Looks great. Thanks @timmc-edx. Mostly a bunch of nits that you could take or leave. I would like to see DLQ acknowledged in the ADR though.

edx_event_bus_kafka/internal/consumer.py
timmc-edx: Ah yeah, it would be hell on the logs, and could be recording hundreds of stack traces per second. :-/ Maybe worth addressing now, then.

My instinct is this: If a message was received, don't worry about tight error loops. If we couldn't get a message _at all_, sleep for a short delay before continuing.

I think this would be relatively uncontroversial, but this approach is complicated by our use of DeserializingConsumer, whose poll call encompasses getting a message, talking to the schema registry, and actually deserializing. If we were using Consumer directly, we could separate out the error cases more easily because the poll call would be on another line, separated from deserialization.

DeserializingConsumer [documents](https://github.com/confluentinc/confluent-kafka-python/blob/395f93901d3296b89bd5b7f9c3de0a2a5a291411/src/confluent_kafka/deserializing_consumer.py#L95) three exceptions it may raise, but looking at the code it's non-exhaustive—I suspect that `Consumer.poll()` can actually raise exceptions in some cases. So perhaps the right thing to do is something like this pseudocode:

```
do_pause = False
try:
    poll()
except KeyDeserializationError, ValueDeserializationError:
    pass
except:
    do_pause = True  # should only be ConsumeError, but could be something unknown

...

# in finally: clause
if do_pause:
    time.sleep(1.0)
```

edx_event_bus_kafka/internal/consumer.py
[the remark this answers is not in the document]
timmc-edx: It's a list of `TopicPartition`: https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.TopicPartition

We should expect to get a list of one of these. (They print nicely.)

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -85,45 +96,54 @@ def inner_from_dict(event_data_dict, ctx=None):  # pylint: disable=unused-argume
...
+                # Allow unit tests to break out of loop
+                if self._shut_down_loop:
+                    break
+
+                msg = None
+                try:
+                    msg = self.consumer.poll(timeout=CONSUMER_POLL_TIMEOUT)
+                    # DeserializingConsumer.poll() always returns either a valid message
+                    # or None, and raises an exception in all other cases. This means
+                    # we don't need to check msg.error() ourselves.
```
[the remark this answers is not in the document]
timmc-edx: It's the same one, yeah. But we're guaranteed `error() is None` by this line: https://github.com/confluentinc/confluent-kafka-python/blob/master/src/confluent_kafka/deserializing_consumer.py#L103

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -135,27 +155,61 @@ def emit_signals_from_message(self, msg):
...
-                f"Signal types do not match. Expected {self.signal.event_type}."
+            raise UnusableMessageError(
+                f"Signal types do not match. Expected {self.signal.event_type}. "
                 f"Received message of type {event_type_str}."
             )
-            return
 
         self.signal.send_event(**msg.value())
 
+    def record_event_consuming_error(self, run_context, error, maybe_msg):
```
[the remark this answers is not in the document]
timmc-edx: I've also just been torn between using "event" and "message" in various places, and found myself using both as synonyms with no particular rhyme or reason. (We're calling it the Event Bus, but in Kafka they're called messages, and this parameter is of type Optional[Message].) But elsewhere in _this method_ at least, it's almost always "event", so I'm thinking `maybe_event`.

I'm definitely more prone to abbreviations for variable names on average. :sweat_smile: I can't find it in the style guide, but I'm OK with being more verbose.

docs/decisions/0009-baseline-error-handling-for-consumer.rst
timmc-edx: Yeah, DLQ was the last thing I felt was missing, but I was still mulling over what to write. Thanks for the suggestions!

edx_event_bus_kafka/internal/tests/test_consumer.py
timmc-edx: Yeah. Although... I've just discovered that `side_effect` can take a *list* of functions to call, which would allow me to get rid of the manual sequencing: https://docs.python.org/3.8/library/unittest.mock.html#unittest.mock.Mock.side_effect

edx_event_bus_kafka/internal/tests/test_consumer.py
timmc-edx: Yeah, maybe `expected_loop_iterations = 4` or something.

edx_event_bus_kafka/internal/tests/test_consumer.py
timmc-edx: Oh... no luck. Emphasis mine:

> If side_effect is an iterable then each call to the mock will **return** the next value from the iterable.

edx_event_bus_kafka/internal/consumer.py
robrap: We'll see if this is a bad call later, but I say we leave this as-is for now. Since we are committing, we should be down to one failure per message. We want to capture the failures in the log anyway. If it is noisy, it means there are lots of messages getting dropped, and hopefully this results in an alert and fixing the issue quickly.  Maybe this will force us to learn about specific errors that should be handled differently? Anyway, just a thought to keep the code clean and to move on and learn from reality.

timmc-edx: I believe we can currently have an arbitrarily large number of errors even with no messages -- it could raise a "couldn't talk to broker" error over and over again in a tight loop.

edx_event_bus_kafka/internal/consumer.py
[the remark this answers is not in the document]
timmc-edx: Yes, true -- that would be safer.

edx_event_bus_kafka/internal/consumer.py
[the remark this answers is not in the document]
timmc-edx: For reference, the assignment would print out as something like this: `[TopicPartition{topic=hello,partition=4,offset=-1001,error=None}]`

I think I'll adjust the message to specifically pull out the partition information since the topic is redundant between the `run_context` and the assignment.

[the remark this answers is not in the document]
timmc-edx: Hmm no, right, assignments is a list and I don't want to deal with "what if the list isn't of size 1 and/or the topic doesn't match..." -- I've pushed something up, see what you think.

docs/decisions/0009-baseline-error-handling-for-consumer.rst
robrap: Nit: Would you mind renaming to "Deferred/Rejected Alternatives"? That would be my preference, given the content, but understood if you feel differently (i.e. are more concerned with consistent headers). I've used my suggestion elsewhere, I think.

robrap (commented):
Thanks. Very close.

edx_event_bus_kafka/internal/consumer.py
timmc-edx: Hmm, that's a thought. It would still sleep in cases where there was a schema mismatch, couldn't talk to schema registry, or bad Avro encoding; it would not sleep for unrecognized signal, other bad headers, or of course for receiver errors.

I could live with it. If we get into a situation where we're in a failure loop and sleeping on each iteration and not making progress, we could make the call to adjust the sleep length via remote config. (I'm not sure what exactly that situation might be—but the important thing is that we'd have an escape valve.) Would have to add it to the runbook, though.

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -85,77 +96,126 @@ def inner_from_dict(event_data_dict, ctx=None):  # pylint: disable=unused-argume
...
 
         Arguments:
             msg (Message): Consumed message.
         """
+        # DeserializingConsumer.poll() always returns either a valid message
+        # or None, and raises an exception in all other cases. This means
+        # we don't need to check msg.error() ourselves. But... check it here
+        # anyway for robustness against code changes.
+        if msg.error() is not None:
+            raise Exception(
```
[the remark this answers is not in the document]
timmc-edx: Hmm, yeah.

docs/decisions/0009-baseline-error-handling-for-consumer.rst
timmc-edx: I'd be OK with renaming it.

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -34,6 +35,13 @@
 
 CONSUMER_POLL_TIMEOUT = getattr(settings, 'EVENT_BUS_KAFKA_CONSUMER_POLL_TIMEOUT', 1.0)
 
+# .. setting_name: POLL_FAILURE_SLEEP
```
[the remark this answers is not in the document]
timmc-edx: Ah, whoops, thanks!

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -34,6 +35,13 @@
...
+# .. setting_default: 1.0
+# .. setting_description: When the consumer fails to retrieve an event from the broker,
```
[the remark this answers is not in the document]
timmc-edx: 1. Currently it doesn't only affect those errors, due to the compromise `msg is None` implementation, so it's a little hard to figure out what to write. Maybe I'll tack on `May also sleep in some other error scenarios.` to expose that situation a little.
2. Yeah, should mention this is in seconds.

edx_event_bus_kafka/internal/consumer.py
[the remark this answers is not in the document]
timmc-edx: From some experimentation and searches, it seems that this assignment can change over time, and will likely not be available before the first poll. So I'll remove that. The `run_context` isn't all that interesting without it, but it could be marginally useful so I guess I'll leave the rest of the line in.

robrap (approved):
Nice. Thanks!

PR #82, opened 2022-11-11
feat: Log and record receiver errors the same way as other errors

timmc-edx:
This will aid in error recovery -- if we need to replay a chunk of the
topic, we'll need to have partition and offset information in the logs.
The thrown exception is given its own class in order to help with building
alerts that can discriminate between different sources of errors (business
logic unique to the consuming IDA vs. more generic kafka/avro issues).

This is part of https://github.com/openedx/event-bus-kafka/issues/62

**Merge checklist:**
- [ ] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,8 @@ Change Log
 Unreleased
 **********
 
+[1.9.0] - 2022-11-15
+********************
 Changed
 =======
 * Log and record receiver errors the same way as other errors (with offset, partition, etc.)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,10 @@ Change Log
 Unreleased
 **********
 
+Changed
+=======
+* Log and record receiver errors the same way as other errors (with offset, partition, etc.)
+
 [1.8.1] - 2022-11-10
 ********************
 Changed
--- a/edx_event_bus_kafka/internal/consumer.py
+++ b/edx_event_bus_kafka/internal/consumer.py
@@ -58,6 +58,12 @@ class UnusableMessageError(Exception):
     """
 
 
+class ReceiverError(Exception):
+    """
+    Indicates that one or more receivers of a signal threw when called.
+    """
+
+
 class KafkaEventConsumer:
     """
     Construct consumer for the given topic, group, and signal. The consumer can then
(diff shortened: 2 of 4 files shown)


CHANGELOG.rst
```diff
@@ -14,6 +14,10 @@ Change Log
 Unreleased
 **********
 
+Changed
```
robrap, line 17: Is there a reason you aren't bumping the version yet?

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -58,6 +58,12 @@ class UnusableMessageError(Exception):
     """
 
 
+class ReceiverError(Exception):
+    """
+    Indicates that one or more receivers of a signal threw when called.
```
robrap, line 63: Nit: Maybe "threw an exception." or "threw an exception when called." ?

robrap (commented):
@timmc-edx: Looks good.

edx_event_bus_kafka/internal/consumer.py
timmc-edx: sure!

CHANGELOG.rst
timmc-edx: I was also going to put in a change to record Kafka error codes and didn't feel like making two releases, but looking into it some more I'm not sure if that's something we should actually ask NR to include in their agent. So I'll go and make this a version bump.

robrap (approved):

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -221,10 +233,11 @@ def report_receiver_errors(self, send_results):
             except AttributeError:
                 receiver_name = str(receiver)
 
+            # The stack traces are already logged by django.dispatcher, so just the error message is fine.
```
robrap, line 236: @timmc-edx: I had the thought that the ReceiverError below could include something like "See separate django.dispatcher log messages for stack traces."

timmc-edx: Added some information to https://2u-internal.atlassian.net/wiki/spaces/AT/pages/234258433/Event+Bus+alerts about the split.

I wouldn't be totally opposed to "(stacktrace elsewhere in logs)" or something equally short. We'd want a unit test to ensure that it stays true, though.

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -221,10 +233,11 @@ def report_receiver_errors(self, send_results):
...
             error_list.append(f"{receiver_name}={response!r}")
 
         if len(error_list) > 0:
-            logger.warning(
+            raise ReceiverError(
                 f"{len(error_list)} receiver(s) out of {len(send_results)} "
                 f"produced errors when handling signal {self.signal}: {', '.join(error_list)}"
             )
```
robrap, line 243: Potential quick win:
(stacktrace elsewhere in logs)
```
            # The stack traces were already logged by django.dispatcher. If this is ever
            #   found not to be the case, update the message and consider adding stacktraces.
            raise ReceiverError(
                f"{len(error_list)} receiver(s) out of {len(send_results)} "
                f"produced errors (stacktrace elsewhere in logs) when handling signal {self.signal}: {', '.join(error_list)}"
            )
```

edx_event_bus_kafka/internal/consumer.py
robrap: If we want to be fast about it, we could consider: https://github.com/openedx/event-bus-kafka/pull/82/files#r1025813178. Otherwise, maybe the test is really fast. Otherwise, we should skip and move on.

timmc-edx: Here's a PR to do this: https://github.com/openedx/event-bus-kafka/pull/84

PR #83, opened 2022-11-15
docs: Update error-handling ADRs; mark as accepted

timmc-edx:
This reflects how our handling of receiver errors changed a bit in commit 0f0874ce51dd/PR #82.

This is part of https://github.com/openedx/event-bus-kafka/issues/62

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/docs/decisions/0009-baseline-error-handling-for-consumer.rst
+++ b/docs/decisions/0009-baseline-error-handling-for-consumer.rst
@@ -4,7 +4,7 @@
 Status
 ******
 
-**Provisional**
+Accepted
 
 Context
 *******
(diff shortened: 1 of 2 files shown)


docs/decisions/0009-baseline-error-handling-for-consumer.rst
```diff
@@ -33,9 +33,9 @@ Consumers of the Event Bus are generally expected to tolerate duplicate events (
 
 As we gain experience, we can add special handling for certain known cases that would be better served by a backoff, retry, or halt.
 
-Receiver exceptions are not *raised*, per se, but are returned by ``send_robust``. We will log these but not treat these as consumer errors, as these exceptions are much more likely to be related to IDA business logic than to the Event Bus itself. There may also be multiple receivers, and failures in one do not indicate that other receivers should stop receiving the event stream.
+Receiver exceptions are not *raised*, per se, but are returned by ``send_robust``. These exceptions are much more likely to be related to IDA business logic than to the Event Bus itself. There may also be multiple receivers, and failures in one do not indicate that other receivers should stop receiving the event stream. We will log these in the same way as other exceptions so that IDA bugs can be recovered from by replaying events, but they'll be specially marked as receiver errors so that alerting can ignore them, if needed.
```
robrap, line 36: Assorted thoughts that might influence some rewording, but I don't have a particular proposal yet:
* I'd clarify that the "IDA business logic" is the consumer event processing logic.
* Noting an assumption of idempotency, it seems likely that if the event is not processed correctly, we would likely want to replay them.
* Noting that these errors will be specially designated seems fine. "Ignoring alerting" seems odd, because presumably someone should know about these failures. Maybe you are trying to say that this enables different teams to alert on different types of errors, but I don't think that is necessary here. Either way, I wouldn't imply the errors might be ignored.

timmc-edx: 1. Open to wording suggestions on the IDA business logic thing.
2. The Consequences section does discuss idempotency, although not using that term.
3. Yeah, I'll rework the bit about "ignoring" -- I did indeed mean that different teams could filter in/out different kinds of errors.

robrap (approved):

PR #87, opened 2022-11-30
feat: add CloudEvent headers to events

rgraber:
Issue: https://github.com/openedx/openedx-events/issues/77

A couple of implementation notes:
event_metadata is left as optional to avoid complicated version management between this, openedx-events, and edx-platform. Once we have everything working in prod we may be able to go back and make it mandatory.

It's unclear to me whether the correct header is "content-type" or "ce_datacontenttype." OEP-41 specifies datacontenttype but the CloudEvent binding for Kafka uses "content-type" in the example. However, it also contains the directive **For the binary mode, the header content-type property MUST be mapped directly to the CloudEvents datacontenttype attribute.**  

The content type is set to "avro" instead of "cloudevents+avro" to prevent any consuming application from trying to parse these as structured content (as opposed to binary content).

### DIFF
--- a/edx_event_bus_kafka/internal/producer.py
+++ b/edx_event_bus_kafka/internal/producer.py
@@ -1,7 +1,7 @@
 """
 Produce Kafka events from signals.
 
-Main function is ``get_producer()``.
+Main function is ``create_producer()``, which should be referred to from ``EVENT_BUS_PRODUCER``.
 """
 
 import json
@@ -17,6 +17,7 @@ from django.conf import settings
 from django.dispatch import receiver
 from django.test.signals import setting_changed
 from edx_django_utils.monitoring import record_exception
+from openedx_events.event_bus import EventBusProducer
 from openedx_events.event_bus.avro.serializer import AvroSignalSerializer
 from openedx_events.tooling import OpenEdxPublicSignal
 
@@ -211,7 +212,7 @@ class ProducingContext:
                         f"partition={evt.partition()}")
 
 
-class KafkaEventProducer():
+class KafkaEventProducer(EventBusProducer):
     """
     API singleton for event production to Kafka.
 
@@ -328,14 +329,9 @@ def poll_indefinitely(api_weakref: KafkaEventProducer):
             api_object = None
 
 
-# Note: This caching is required, since otherwise the Producer will
-# fall out of scope and be garbage-collected, destroying the
-# outbound-message queue and threads. The use of this cache allows the
-# producer to be long-lived.
-@lru_cache  # will just be one cache entry, in practice
-def get_producer() -> Optional[KafkaEventProducer]:
+def create_producer() -> Optional[KafkaEventProducer]:
     """
-    Create or retrieve Producer API singleton.
+    Create a Producer API instance. Caller should cache the returned object.
 
     If confluent-kafka library or essential settings are missing, warn and return None.
     """
@@ -354,4 +350,3 @@ def get_producer() -> Optional[KafkaEventProducer]:
 def _reset_caches(sender, **kwargs):  # pylint: disable=unused-argument
     """Reset caches when settings change during unit tests."""
     get_serializers.cache_clear()
-    get_producer.cache_clear()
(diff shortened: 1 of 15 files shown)


edx_event_bus_kafka/internal/producer.py
```diff
@@ -177,18 +183,44 @@ def inner_to_dict(event_data, ctx=None):  # pylint: disable=unused-argument
...
+    Create a dictionary of CloudEvent-compliant Kafka headers from an EventsMetadata object
+
+    Arguments:
+        event_metadata: An EventsMetadata object sent by an OpenEdxPublicSignal
+
+    Returns:
+        A dictionary of headers
+    """
+    if not event_metadata:
+        return {EVENT_TYPE_HEADER_KEY: signal.event_type.encode("utf-8")}
```
robrap, line 197: 1. I don't love the duplicated code.
2. I'm not clear on why the other non-metadata headers wouldn't always be included. I can see maybe leaving off specversion unless all CloudEvent required fields are there? But, do we want to return a partial header, or should we just be returning an error if it isn't provided?
```
header = {}

if event_metadata:
    header = {
        ...
    }

header[EVENT_TYPE_HEADER_KEY] = signal.event_type.encode("utf-8")
header[SPEC_VERSION_HEADER_KEY] = ...
header[CONTENT_TYPE_HEADER_KEY] = ...

return header
```

robrap (commented):
I was not very thorough on the test review. I hope that is ok. :)

robrap:
@rgraber: I agree that I was confused on `content-type`. As I noted on the issue, what if we just create `ce_datacontenttype` as a redundant header? No matter what way we go, we might want to add a code comment about this where we are setting it.

edx_event_bus_kafka/internal/producer.py
```diff
@@ -245,22 +279,19 @@ def send(
             event_key_field: Path to the event data field to use as the event key (period-delimited
               string naming the dictionary keys to descend)
             event_data: The event data (kwargs) sent to the signal
+            event_metadata: An EventsMetadata object with all the metadata necessary for the CloudEvent spec
```
robrap, line 282: Should we mention in code that this is optional only for rollout purposes?

edx_event_bus_kafka/internal/producer.py
rgraber: I mostly did this for rollout purposes since I figured we'd probably want to make event_metadata mandatory. If we return an error when we don't have event_metadata we're setting ourselves up for a bunch of errors from studio if we don't upgrade very carefully.

edx_event_bus_kafka/internal/producer.py
```diff
@@ -177,18 +183,44 @@ def inner_to_dict(event_data, ctx=None):  # pylint: disable=unused-argument
...
+
+    # Dictionary (or list of key/value tuples) where keys are strings and values are binary.
+    # CloudEvents specifies using UTF-8; that should be the default, but let's make it explicit.
+    return {
+        EVENT_TYPE_HEADER_KEY: signal.event_type.encode("utf-8"),
+        ID_HEADER_KEY: str(event_metadata.id).encode("utf-8"),
+        SOURCE_HEADER_KEY: event_metadata.source.encode("utf-8"),
+        SOURCEHOST_HEADER_KEY: event_metadata.sourcehost.encode("utf-8"),
+        # Always 1.0. See "Fields" in OEP-41
+        SPEC_VERSION_HEADER_KEY: b'1.0',
```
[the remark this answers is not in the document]
rgraber: oh good catch, that was a mistake on my part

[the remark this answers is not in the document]
rgraber: [fixed]

edx_event_bus_kafka/internal/producer.py

edx_event_bus_kafka/internal/producer.py
rgraber: [wontfix unless you want to argue about it]

robrap: Ok. If you see fit, you could copy the comment here as well:
```
# TODO: Make event_metadata required (https://github.com/openedx/openedx-events/issues/153)
```
Not a big deal either way, as long as you remember in the next ticket. Or, we could just add a link to this thread in that ticket. Or, you can just remember.

robrap (approved):

PR #89, opened 2022-12-02
feat!: make event_metadata a required parameter of send()

rgraber:
**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [N/A] Commits are squashed
- [N/A] Noted any: Concerns, dependencies, deadlines, tickets, testing instructions

### DIFF
--- a/edx_event_bus_kafka/internal/producer.py
+++ b/edx_event_bus_kafka/internal/producer.py
@@ -186,7 +186,7 @@ def get_serializers(signal: OpenEdxPublicSignal, event_key_field: str):
     return key_serializer, value_serializer
 
 
-def get_headers_from_metadata(event_metadata: EventsMetadata):
+def _get_headers_from_metadata(event_metadata: EventsMetadata):
     """
     Create a dictionary of CloudEvent-compliant Kafka headers from an EventsMetadata object.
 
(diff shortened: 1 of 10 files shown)


edx_event_bus_kafka/internal/producer.py
```diff
@@ -186,24 +186,24 @@ def inner_to_dict(event_data, ctx=None):  # pylint: disable=unused-argument
     return key_serializer, value_serializer
 
 
-def get_headers_from_signal_and_metadata(signal: OpenEdxPublicSignal, event_metadata: EventsMetadata):
+def get_headers_from_metadata(event_metadata: EventsMetadata):
```
robrap, line 189: Is this private? Can we add back signal when and if we need it? I really like using underscores, like `_get_headers_from_metadata`, but maybe we aren't doing that and it would be a bigger change. It just makes it so much nicer for others (like me) to come into code and no what is simpler to refactor, and what is changing an API.

robrap (approved):

edx_event_bus_kafka/internal/producer.py

rgraber: jk, making this private makes it hard to test, and since it's outside the producer class proper hopefully it's clear that it's not part of the API. Thoughts?

PR #92, opened 2022-12-08
feat: Add the ability to override offsets.

dianakhuang:
In the case where we would want to replay events,
we would need to change the offsets in the partitions so that we can consume from an earlier point. This allows users to pass in a new offset parameter that will reset the offsets before starting to consume again.

https://github.com/edx/edx-arch-experiments/issues/115

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 **********
 
+[3.2.0] - 2022-12-14
+********************
+Changed
+=======
+* Add timestamp parameter for consumer, allowing the starting offset for consuming to be overridden from the default.
+
 [3.1.0] - 2022-12-07
 ********************
 
--- a/edx_event_bus_kafka/internal/consumer.py
+++ b/edx_event_bus_kafka/internal/consumer.py
@@ -4,6 +4,7 @@ Core consumer and event-loop code.
 
 import logging
 import time
+from datetime import datetime
 
 from django.conf import settings
 from django.core.management.base import BaseCommand
@@ -19,7 +20,7 @@ logger = logging.getLogger(__name__)
 # See https://github.com/openedx/event-bus-kafka/blob/main/docs/decisions/0005-optional-import-of-confluent-kafka.rst
 try:
     import confluent_kafka
-    from confluent_kafka import DeserializingConsumer
+    from confluent_kafka import DeserializingConsumer, TopicPartition
     from confluent_kafka.error import KafkaError
     from confluent_kafka.schema_registry.avro import AvroDeserializer
 except ImportError:  # pragma: no cover
@@ -131,10 +132,36 @@ class KafkaEventConsumer:
         """
         self._shut_down_loop = True
 
-    def consume_indefinitely(self):
+    def consume_indefinitely(self, offset_timestamp=None):
         """
         Consume events from a topic in an infinite loop.
+
+        Arguments:
+            offset_timestamp(datetime): reset the offsets of the consumer partitions to this timestamp before consuming.
         """
+
+        def reset_offsets(consumer, partitions):
+            # This is a callback method used on consumer assignment to handle offset reset logic.
+            # We do not want to attempt to change offsets if the offset is None.
+            if offset_timestamp is None:
+                return
+
+            # Get the offset from the epoch in seconds.
+            offset_timestamp_s = int(offset_timestamp.timestamp())
+            # We construct partitions with the epoch timestamp in the offset position.
+            timestamp_partitions = [TopicPartition(self.topic, p.partition, offset_timestamp_s) for p in partitions]
+
+            partitions_with_offsets = consumer.offsets_for_times(timestamp_partitions, timeout=1.0)
+
+            # Partitions have an error field that may be set on return.
+            errors = [p.error for p in partitions_with_offsets if p.error is not None]
+            if len(errors) > 0:
+                raise Exception("Error getting offsets for timestamps: {errors}")
+
+            logger.info(f'Found offsets for timestamp {offset_timestamp}: {partitions_with_offsets}')
+
+            consumer.assign(partitions_with_offsets)
+
         # This is already checked at the Command level, but it's possible this loop
         # could get called some other way, so check it here too.
         if not KAFKA_CONSUMERS_ENABLED.is_enabled():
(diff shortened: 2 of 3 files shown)


edx_event_bus_kafka/internal/consumer.py
```diff
@@ -133,6 +134,16 @@ def consume_indefinitely(self):
         """
         Consume events from a topic in an infinite loop.
         """
+
+        def reset_offsets(consumer, partitions):
+            # This is a callback method used on consumer assignment to handle offset reset logic.
+            if self.offset is None:
+                return
+            for partition in partitions:
```
rgraber, line 142: Aren't we going to want different offsets for each partition?

rgraber (commented):
I think we need a partition:offset map since different partitions will need to be replayed from different offsets

edx_event_bus_kafka/internal/consumer.py
dianakhuang: hmm, probably. I think it would be easier then to use timestamps instead of offset values.

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -133,6 +134,20 @@ def consume_indefinitely(self):
...
+
+        def reset_offsets(consumer, partitions):
+            # This is a callback method used on consumer assignment to handle offset reset logic.
+            # We do not want to attempt to change offsets if the offsets is None or if it's an empty list.
+            if not self.offsets:
+                return
+            if len(self.offsets) != len(partitions):
+                logger.error("Offsets given do not match the number of partitions. Skipping reset step.")
+                return
+            for partition, offset in zip(partitions, self.offsets):
```
timmc-edx, line 146: I don't think we're guaranteed which partitions we'll be assigned, or in what order.

dianakhuang: yeah, this code was very much of the 'let's see what it would look like to pass in a list of offsets', but since we're not going this route, i don't think we have to worry about it.

CHANGELOG.rst
```diff
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 **********
 
+[3.1.0] - 2022-12-08
+********************
+Changed
+=======
+* Add offset parameter for consumer, allowing the starting offset for consuming to be overridden from default.
```
timmc-edx, line 21: Should be adjusted a bit to account for timestamp-based approach.

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -322,17 +339,34 @@ def add_arguments(self, parser):
...
+                    offset_timestamp = int(datetime.fromisoformat(options['offset_time'][0]).timestamp())
+                except ValueError:
+                    logger.exception('Could not parse the offset timestamp.')
+                    raise
+
             event_consumer = KafkaEventConsumer(
                 topic=options['topic'][0],
                 group_id=options['group_id'][0],
                 signal=signal,
+                offset_timestamp=offset_timestamp,
```
timmc-edx, line 369: Consider passing this param to `consume_indefinitely` directly instead?

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -322,17 +339,33 @@ def add_arguments(self, parser):
...
 
     def handle(self, *args, **options):
         if not KAFKA_CONSUMERS_ENABLED.is_enabled():
             logger.error("Kafka consumers not enabled, exiting.")
             return
         try:
             signal = OpenEdxPublicSignal.get_signal_by_type(options['signal'][0])
+            if options['offset_time'][0] is not None:
+                try:
+                    offset_timestamp = int(datetime.fromisoformat(options['offset_time'][0]).timestamp())
```
timmc-edx, line 359: It might be safest to do the parsing here, but pass it around as a datetime object -- then convert it to seconds or milliseconds from epoch only at the place where we'll use it. (Alternatively, a name like `offset_timestamp_s` to indicate that it's denominated in seconds.)

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -129,10 +130,30 @@ def _shut_down(self):
...
+            # We do not want to attempt to change offsets if the offset is None.
+            if offset_timestamp is None:
+                return
+
+            # Get the offset from the epoch in seconds.
+            offset_timestamp_s = int(offset_timestamp.timestamp())
+            # We construct partitions with the epoch timestamp in the offset position.
+            timestamp_partitions = [TopicPartition(self.topic, p.partition, offset_timestamp_s) for p in partitions]
+
+            partitions_with_offsets = consumer.offsets_for_times(timestamp_partitions, timeout=1.0)
```
timmc-edx, line 152: I believe we'll also need to check the `error` field on the partitions: https://github.com/confluentinc/confluent-kafka-python/pull/268/files#diff-34dc92cd663b144203a9252b1eeb69bc3f10cdaaea64f9ec379503be3cc62485R1001

> returns: list of topic+partition with offset field set and possibly error set

dianakhuang: Not to kick the can too far down the road, but I think this might also be handled by Kafka itself when it looks through the list of partitions?

timmc-edx: Hmm... that would be good, but I can't find anything reassuring in confluent-kafka or librdkafka in that direction. Here's librdkafka's implementation of assign: https://github.com/edenhill/librdkafka/blob/v1.9.2/src/rdkafka_subscription.c#L119

It's possible we'd get an error during poll or something else, but I'd rather stay on the safe side and check that all the error attributes are None. Otherwise we might end up with something goofy like _actually_ having timestamps in the offsets, etc.

dianakhuang: Okay, will add some error handling around this, then.

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -129,10 +130,38 @@ def _shut_down(self):
...
+
+            # Partitions have an error field that may be set on return.
+            errors = [p.error for p in partitions_with_offsets if p.error is not None]
+            if len(errors) > 0:
+                for error in errors:
+                    logger.error(f"Error getting offsets for timestamps: {error}")
+                raise Exception("Error getting offsets for timestamps")
```
timmc-edx, line 159: Maybe just collapse it down to this?
```suggestion
                raise Exception("Error getting offsets for timestamps: {errors}")
```

timmc-edx (approved):
Looking good!

(It's unfortunate that we don't have a good way to unit test, but I don't have any bright ideas.)

PR #105, opened 2023-01-03
feat: Allow exiting consumer if too many errors in a row

timmc-edx:
This is intended to help with a stuck DB connection issue we observed:
https://github.com/edx/edx-arch-experiments/issues/144

PR is divided into two commits; first commit is a small testing refactor (extraction of a helper).

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/edx_event_bus_kafka/internal/consumer.py
+++ b/edx_event_bus_kafka/internal/consumer.py
@@ -169,6 +169,20 @@ class KafkaEventConsumer:
             logger.error("Kafka consumers not enabled, exiting.")
             return
 
+        # .. setting_name: EVENT_BUS_KAFKA_CONSUMER_CONSECUTIVE_ERRORS_LIMIT
+        # .. setting_default: None
+        # .. setting_description: If the consumer encounters this many consecutive errors, exit with an
+        #   error. This is intended to be used in a context where a management system (such as Kubernetes)
+        #   will relaunch the consumer automatically. The effect is that all runtime state is cleared,
+        #   allowing consumers in arbitrary "stuck" states to resume their work automatically. (The impetus
+        #   for this setting was a Django DB connection failure that was staying failed.) Process
+        #   managers like Kubernetes will use delays and backoffs, so this may also help with transient
+        #   issues such as networking problems or a burst of errors in a downstream service. Errors may
+        #   include failure to poll, failure to decode events, or errors returned by signal handlers.
+        #   This does not prevent committing of offsets back to the broker; any messages that caused an
+        #   error will still be marked as consumed, and may need to be replayed.
+        CONSECUTIVE_ERRORS_LIMIT = getattr(settings, 'EVENT_BUS_KAFKA_CONSUMER_CONSECUTIVE_ERRORS_LIMIT', None)
+
         try:
             full_topic = get_full_topic(self.topic)
             run_context = {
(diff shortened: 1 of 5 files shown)


edx_event_bus_kafka/internal/consumer.py
```diff
@@ -169,6 +169,19 @@ def reset_offsets(consumer, partitions):
             logger.error("Kafka consumers not enabled, exiting.")
             return
 
+        # .. setting_name: EVENT_BUS_KAFKA_CONSUMER_CONSECUTIVE_ERRORS_LIMIT
+        # .. setting_default: None
+        # .. setting_description: If the consumer encounters this many consecutive errors, exit with an
```
robrap, line 174: First, I was wondering if we should document the lost connection issue, since this is what we were proposing as the solution. Next, I was wondering what database config solutions there might be, and whether that was looked into, for those that don't want to auto-restart.

Here are the latest Django docs for db: https://docs.djangoproject.com/en/4.1/ref/databases/#persistent-connections. An interesting quote:
> The default value is 0, preserving the historical behavior of closing the database connection at the end of each request. To enable persistent connections, set [CONN_MAX_AGE](https://docs.djangoproject.com/en/4.1/ref/settings/#std-setting-CONN_MAX_AGE) to a positive integer of seconds. For unlimited persistent connections, set it to None.

This got me wondering if we should just force close the connection, which got me to this stackoverlow answer. Not sure if `connect()` is correct, but it could be. See https://stackoverflow.com/a/50694172/13500961

Not sure if we want both these changes (fix db connection and exit after error limit).

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -179,18 +192,32 @@ def reset_offsets(consumer, partitions):
             self.consumer.subscribe([full_topic], on_assign=reset_offsets)
             logger.info(f"Running consumer for {run_context!r}")
 
+            # How many errors have we seen in a row? If this climbs too high, exit with error.
+            # Any error counts, here—whether due to a polling failure or a message processing
```
robrap, line 196: Assuming `here-whether` is a typo?

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -179,18 +192,32 @@ def reset_offsets(consumer, partitions):
...
+            # being able to talk to the broker and get a message (or a normal poll timeout) is
+            # not sufficient to show that progress can be made.
+            consecutive_errors = 0
+
             while True:
                 # Allow unit tests to break out of loop
                 if self._shut_down_loop:
                     break
 
+                # Detect probably-broken consumer and exit with error.
```
robrap, line 207: Nit: My brain would put this at the end of the loop, but I guess yours wouldn't. :)

robrap (commented):
Looks good. Ping me if you want to discuss the stuck connection specifics.

edx_event_bus_kafka/internal/consumer.py
timmc-edx: DB connection config would be outside of the scope of event-bus-kafka, but we may want to mention it in our Event Bus docs. And certainly we should consider trying that `connect()` change in the Discovery workers.

edx_event_bus_kafka/internal/consumer.py
timmc-edx: No—it's an em-dash, but I guess those don't show up well in fixed-width fonts. :-)

edx_event_bus_kafka/internal/consumer.py
timmc-edx: I remember considering whether I wanted it at the beginning or the end, but I don't remember why I chose this. 🤔 I suspect it's because of the general principle that there can be multiple ways to exit an iteration (`continue` or falling through), but only one way to enter one. There _isn't_ a `continue` statement here, so it could go at the end as well. The other likely reason is that I wanted it near the `_shut_down_loop`, thematically. :-) But neither is a particularly overwhelming reason.

edx_event_bus_kafka/internal/consumer.py
robrap: Since we're implementing with Django, and implementing an infinite loop (which is different than a request based model), it feels like we'd want to provide reasonable behavior by default if we could. That said, I'm not sure what "reasonable behavior" should be. We have `CONN_MAX_AGE` set to 60, so the above comment shouldn't apply. But, I'm not sure what should happen, because our connection stayed broken beyond 1 minute. Does that mean it requires a `connect()` call to fix the connection, and that isn't happening without a new request?

When one writes a new Django view, this is not something they need to think about, whether the view uses or doesn't use the database. How does that work, and could we replicate that behavior by default (mimicking what Django does)? I don't like the idea that every new signal handler has to worry about this.

robrap: This being said, I'm now not sure how to best document or fix from a db connection perspective, so maybe we just land this and ticket the rest as an open issue that can sit in future, given that we have this backstop.

robrap (approved):

edx_event_bus_kafka/internal/consumer.py
timmc-edx: Hmm... yeah, a close reading of the docs there indicates that the max age is only checked at request start, and usability is only checked at the end of the request, and (I think) a new connection is always only provided at the start of a request, in any case.

I think what we want to do is try various approaches in the discovery-worker management command code, and then determine if the fix can be lifted out—and to where. First I'd just like to get this generic backstop out there, though.

PR #106, opened 2023-01-06
fix: pass offsets to partitions as milliseconds instead of seconds

rgraber:
Kafka partitions expect time offsets to be in milliseconds instead of seconds, so multiply the timestamp by 1000 before setting it on the partition.

Tested locally against the dev cluster.

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [ ] Commits are squashed


edx_event_bus_kafka/internal/consumer.py
```diff
@@ -146,11 +146,12 @@ def reset_offsets(consumer, partitions):
             if offset_timestamp is None:
                 return
 
-            # Get the offset from the epoch in seconds.
-            offset_timestamp_s = int(offset_timestamp.timestamp())
+            # Get the offset from the epoch (Kafka expects offsets in milliseconds for offsets_for_times, see
+            # https://kafka-python.readthedocs.io/en/master/apidoc/KafkaConsumer.html#kafka.KafkaConsumer.offsets_for_times)
```
timmc-edx, line 150: Might be misleading to link to the kafka-python library's docs here, since we're not using it.

rgraber: Good catch. I'll see if I can find it in the library we're actually using.

timmc-edx: As far as I can tell it's undocumented, both for confluent-kafka and the underlying librdkafka. I have a PR open to add a note about it in the latter.

CHANGELOG.rst
```diff
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 **********
 
+[3.5.1] - 2023-01-05
```
timmc-edx, line 17:
```suggestion
[3.5.1] - 2023-01-06
```

edx_event_bus_kafka/internal/consumer.py
timmc-edx, line 150: Maybe something along these lines? Anyway, I'll go ahead and approve, and you can decide what makes the most sense docs-wise.

```suggestion
            # Get the offset from the epoch. Kafka expects offsets in milliseconds for offsets_for_times. Although
            # this is undocumented in the libraries we're using (confluent-kafka and librdkafa), for reference
            # see the docs for kafka-python:
            # https://kafka-python.readthedocs.io/en/master/apidoc/KafkaConsumer.html#kafka.KafkaConsumer.offsets_for_times
```

timmc-edx (approved):

PR #112, opened 2023-01-20
fix: Added poll call to reset-offset code.

dianakhuang:
We believe that the offsets are not reset and assigned until poll is called. This poll should not process events or commit the offsets with the messages it returns.

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [ ] Commits are squashed

### DIFF
--- a/edx_event_bus_kafka/internal/consumer.py
+++ b/edx_event_bus_kafka/internal/consumer.py
@@ -213,6 +213,13 @@ class KafkaEventConsumer:
                     # This log message may be noisy when we are replaying, but hopefully we only see it
                     # once every 30 seconds.
                     logger.info("Offsets are being reset. Sleeping instead of consuming events.")
+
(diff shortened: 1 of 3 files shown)


edx_event_bus_kafka/internal/consumer.py
```diff
@@ -213,6 +213,13 @@ def reset_offsets(consumer, partitions):
                     # This log message may be noisy when we are replaying, but hopefully we only see it
                     # once every 30 seconds.
                     logger.info("Offsets are being reset. Sleeping instead of consuming events.")
+
+                    # We are calling poll here because we believe the offsets will not be set
+                    # correctly until poll is called, despite the offsets being reset in a different call.
```
timmc-edx, line 218: Maybe add something in here about how we don't think partitions are assigned until poll is called?

timmc-edx (approved):

PR #113, opened 2023-01-23
fix: reset db connection after connection is dropped

rgraber:
Issue https://github.com/edx/edx-arch-experiments/issues/158

The event bus consumer workers have a tendency to fail on an ``django.db.utils.OperationalError: (2013, 'Lost connection to MySQL server during query')`` error. Locally, this seems to happen when the mysql container goes down and comes back up while the consume loop is running. Presumably, something similar happens on prod/stage when a database node restarts. 

Even though the database has restarted, the worker is still holding on to an old connection, which has since been closed. We can tell if our connection has been closed with `connection.is_usable`. If the connection is no longer usable, calling `connection.connect()` will get a fresh one.

Documentation is a little confusing, but local testing using `show status where `variable_name` = 'Threads_connected';` in the local mysql container seems to verify that this does not cause a connection leak. Also the connection is successfully closed when the management command is forced to exit.

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [ ] Commits are squashed


edx_event_bus_kafka/internal/consumer.py
```diff
@@ -132,6 +138,39 @@ def _shut_down(self):
...
+        Returns:
+            True if we want to commit the message offset, False if not
+            Raises if the error is a Kafka fatal error
+        """
+        # case 1: DB error - reconnect and try again by rewinding the offset
+        if isinstance(e, OperationalError):
+            error_code, _ = e.args
+            if error_code == MYSQL_LOST_CONNECTION_ERROR:
+                message_location = TopicPartition(msg.topic(), msg.partition(), msg.offset())
+                db.connection.close()
```
robrap, line 154: Before doing a detailed review, I'm wondering what happens if we call `db.connection.connect()` at the start of each loop? Does that work? Does it prevent the initial failure? 

Note: if we do this, we'd want to first check that there is a connection defined. We could have a consumer with no db. Also, unless we connect on all connections, we'd need to document that the same should be done in the consumer when using non-default connections, like the read-only db.

rgraber: What do you mean connect on all connections?

robrap: You can ignore the all-connections note. It was based on something I had read, but can’t find, and it would be a more exceptional case anyway.

For now, I’m wondering if `connect()` are the start of each loop works more similarly to Django requests, and fixes the connection as-needed.

rgraber: Scouring the internet and django docs, I can't find any examples of manually calling connect(), which makes me pretty hesitant to use it. We'd also have to manually handle closing it (I think) which opens us up to future possible connection leaks.

robrap: 1. Does it seem to resolve the issue? If not, it isn't worth looking more deeply. If so, it is worth additional consideration.
2. Maybe we could call connect only when we think it is already connected, to refresh the connection as-needed.
3. Is it possible that we already aren't closing connections properly when exiting the loop? Is that something we should be doing? Or is Django treating this as one long "request", and when we exit, it is handling all of its usual clean-up activities?

rgraber: For 3, I don't know specifics, but I can't imagine Django doesn't have built-in tools for managing the connection for a management command. Most of our management commands talk to the DB so we would notice if those weren't getting closed properly.

rgraber: I know Django will autoclose connections based on a configured timeout

robrap: I'm not too worried about 3, but I will point out that most management commands end, and ours runs indefinitely, so there could be django clean-up code that never gets run.

edx_event_bus_kafka/internal/tests/test_consumer.py
```diff
@@ -305,6 +305,41 @@ def raise_exception():
...
+        """Confirm we reconnect to the database if our connection is deemed unusable"""
+        mock_emit = Mock(side_effect=side_effects([self.event_consumer._shut_down]))  # pylint: disable=protected-access
+        with patch.multiple(self.event_consumer,
+                            emit_signals_from_message=mock_emit,
+                            **{'_db_needs_reconnect': Mock(return_value=True), '_reconnect_to_db': DEFAULT}
+                            ) as mock_event_consumer:
+            mock_consumer = Mock(**{'poll.return_value': self.normal_message}, autospec=True)
+            self.event_consumer.consumer = mock_consumer
+            self.event_consumer.consume_indefinitely()
+        mock_event_consumer['_reconnect_to_db'].assert_called_once()
```
robrap, line 323: 1. I'm used to seeing something like: `mock_event_consumer._reconnect_to_db.assert_called_once()`. Does either work, or am I just not getting this code at all? I didn't even know you could get the method by key, but maybe I'm missing something and this is different from what I think.
2. What does `DEFAULT` do?

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -132,6 +133,24 @@ def _shut_down(self):
...
 
+    def _db_needs_reconnect(self):
+        """
+        Determine if we need to try to reconnect to the database
+
+        Returns:
+            True if we have a db connection but it is unusable
+            False otherwise (if we don't have a connection or our current connection is fine)
+        """
+        return connection.connection and not connection.is_usable()
```
robrap, line 144: 1. Checking `connection.connection` first might be totally necessary, but seems really odd. Should we link to docs? Does it raise an exception if you call `connection.is_usable()` without checking this? If it isn't needed, I think we should drop it.
2. [Nit] If it is needed, I wonder if the following would make it a little more readable:
```
has_connection = bool(connection.connection)
return has_connection and not connection.is_usable()
```

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -132,6 +133,24 @@ def _shut_down(self):
...
+
+    def _reconnect_to_db(self):
+        """
+        Pass-through to connection.connect()
+
+        Moved out for easier testing (mocking connections in pytest is hard)
```
robrap, line 150: [Nit-ish] Mocking is already once removed from reality, and mocking the wrappers seems even further removed. If you simply just patch/mock the top-level `connection`, can you use code like the following?:
```
mock_connection.connection = None
mock_connection.connect.return_value = True
mock_connection.is_usable.return_value = True
etc.
```

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -132,6 +133,24 @@ def _shut_down(self):
...
+
+        Returns:
+            True if we have a db connection but it is unusable
+            False otherwise (if we don't have a connection or our current connection is fine)
+        """
+        return connection.connection and not connection.is_usable()
+
+    def _reconnect_to_db(self):
+        """
+        Pass-through to connection.connect()
```
robrap, line 148: What about something like:
> Used to re-connect broken connections by calling connection.connect()

Note: even if we change our mocking (see other comment), you may still want this wrapper to have a better name and docstring. But, maybe it is overkill. Not sure.

edx_event_bus_kafka/internal/tests/test_consumer.py
```diff
@@ -305,6 +305,41 @@ def raise_exception():
...
 
+    @override_settings(
+        EVENT_BUS_KAFKA_SCHEMA_REGISTRY_URL='http://localhost:12345',
+        EVENT_BUS_KAFKA_BOOTSTRAP_SERVERS='localhost:54321',
+        EVENT_BUS_TOPIC_PREFIX='dev',
+    )
+    def test_connection_reset_if_unusable(self):
+        """Confirm we reconnect to the database if our connection is deemed unusable"""
+        mock_emit = Mock(side_effect=side_effects([self.event_consumer._shut_down]))  # pylint: disable=protected-access
+        with patch.multiple(self.event_consumer,
```
robrap, line 316: I get so lost in this. :) Is it possible to just patch what we need in a decorator on the method, and leave this complicated code for breaking out of the loop alone, so it doesn't get even more complicated?

robrap (commented):
Awesome! Minor comments about coding/mocking, but so glad you got this fixed.

robrap:
I think we need a doc (or section of the how-to, which may exist) about the abstraction layer. Any new implementation should copy this pattern, unless and until we break out reusable code from the consumer loop. We need to document this (at a minimum). We could extend that doc with any other notes about the abstraction layer, and links to the future abstraction tickets.
FYI: @bmtcril

edx_event_bus_kafka/internal/consumer.py
rgraber: `connection.is_usable()` does indeed raise if connection.connection is null. I'm pretty ambivalent as to the structure so will use your suggestion.

edx_event_bus_kafka/internal/consumer.py
rgraber: Mocking the top level connection turns out to wreak havoc on the tests, which is why I separated these wrappers out. I can try again but I don't think it's worth the development hours.

edx_event_bus_kafka/internal/consumer.py
rgraber: I like calling out that it's a straight up passthrough, but maybe I'll put that in the second line

edx_event_bus_kafka/internal/tests/test_consumer.py
rgraber: [Fixed]  Refactored to not use `patch.multiple`.

[inform] getting the method by key is how  the return value of `patch.multiple()` works. DEFAULT just means use a default mock (no custom return method or anything).

edx_event_bus_kafka/internal/consumer.py
rgraber: [fixed]

edx_event_bus_kafka/internal/consumer.py
rgraber: [fixed] although I kept my line as the more detailed explanation

edx_event_bus_kafka/internal/tests/test_consumer.py
rgraber: [fixed]

edx_event_bus_kafka/internal/consumer.py
rgraber: [wontfix]

robrap: I hope you don't mind, but I made a PR for this: https://github.com/openedx/event-bus-kafka/pull/117

robrap:
> I would put that in a new issue. I don't think we want it tied to this bugfix if we need to revert.

I totally agree. Sorry I didn't make that more clear. Will you or did you create this ticket @rgraber?

robrap (approved):
Squash away.

rgraber:
> > I would put that in a new issue. I don't think we want it tied to this bugfix if we need to revert.
> 
> I totally agree. Sorry I didn't make that more clear. Will you or did you create this ticket @rgraber?

Already created. https://github.com/openedx/event-bus-kafka/issues/116

Issue #111, opened 2023-01-18
Add audit logging for event consumption


timmc-edx:
We currently don't have a complete record of when we consumed an event from the broker and then how long it took to process. These would be important information for debugging consumer latency and other issues; right now it's hard to get a picture of how e.g. a slow-to-process event affects the latency of following events, or whether there is a large gap in incoming events that correlates with DB lost connection errors.

New Relic has some information, but due to sampling we don't have a complete picture. (Usually we record information for 99+% of events, but over a 1 month window I see average capture drop as low as 70%. This may be due to heavier loss during replays.) Our logs also have some information, but only for errors.

We'll start by logging all received and processed events -- not necessarily in full detail, but enough for debugging. If at some point we're generating enough log data that it's a problem, we can add feature toggles or otherwise reassess our approach.

A/C:

- [x] Log metadata when an event is received: Key, topic, partition, offset, timestamp when received, production timestamp (if available)
- [x] Log when an event is done processing: Key, topic, partition, offset, timestamp, receiver output. May need to unify with error handler.
- [x] Log when and event is produced with: Key, topic, timestamp 
- [x] Small ADR explaining the need for audit logging

robrap:
Thanks for the ticket.
1. Are there any other bits of metadata we want to add in or leave out? See https://github.com/openedx/openedx-events/blob/main/openedx_events/data.py#L25-L32
2. Do we want any audit from the produce callback?
3. If so, do we want to drop any redundant metadata from the producer?
4. Do we want to add any other consumer-side metadata like the proposed timestamp (e.g. library version)?

robrap:
[question] What do we have in Datadog (from Confluent Kafka metrics) at this time?

robrap:
1. New Relic does contain (sampled?) transactions for each message and its duration. See https://onenr.io/0vwBpDmz8jp. We could pull the offset and partition and count to understand how much sampling.
2. The main point of this ticket is to add additional audit details in the logs if samplings proves to lose too much data to debug effectively.

robrap:
From @timmc-edx:
> [inform] To answer the question of whether NR is sampling Kafka information, I ran this query: from Transaction select max(kafka_offset) - min(kafka_offset), count(*) where appName = 'prod-edx-discovery' facet kafka_partition since 2 days ago Turns out it actually matches up, with the exception of what is probably just an off-by-one issue. Out to 14 days, there might be a couple of missing events, but at least 99.9% of them are there.

[...]

PR #123, opened 2023-01-31
feat: Add audit logging for message produce and consume

timmc-edx:
Addresses https://github.com/openedx/event-bus-kafka/issues/111

- Log identifying information when a message is received by the consumer
- Log on the consumer when a message is successfully processed (not just when there's an error)
- Add offset to event-produced log message

Also:

- Adjust producer logging a bit to match other code
- Add topic and timestamp to FakeMessage (needed for new tests)
- Change env prefix used in unit tests from `prod` to `local` (it makes me nervous to use prod here)
- Remove some unused setting overrides

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed


docs/decisions/0010_audit_logging.rst
```diff
@@ -0,0 +1,36 @@
...
+********
+
+We will log at least minimal information about all Kafka messages that we produce and consume. These log messages will not necessarily have sufficient information to reconstruct a message, but will identify messages uniquely (topic, partition, and offset) and include timestamps that are necessary for reconstructing a timeline of activity. Beyond that we will probably just include identifying information such as message keys and unique IDs, as they are useful for searches but should not add too much chaff to the logs.
+
+Timestamps will be recorded implicitly, via the logging framework's own log formatting.
+
+Consequences
+************
+
+Log volume will be increased, likely not by too much.
```
robrap, line 31: Thoughts on adding a toggle to enable/disable audit logging?

docs/decisions/0010_audit_logging.rst
```diff
@@ -0,0 +1,36 @@
...
+
+Rejected Alternatives
+*********************
+
+N/A
```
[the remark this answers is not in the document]
timmc-edx: Good call!

docs/decisions/0010_audit_logging.rst
timmc-edx: I suppose it might make sense, but I'd be inclined to hold off until it seems necessary. Were you thinking default-on or default-off?

robrap: I was thinking default-on, but either way. Just a kill switch if needed.

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -392,6 +402,32 @@ def _check_receiver_results(self, send_results: list):
...
+        """
+        try:
+            (ts_type, timestamp_ms) = msg.timestamp()
+            if ts_type == TIMESTAMP_NOT_AVAILABLE:
+                timestamp_info = "none"
+            else:
+                # Could be produce time or broker receive time; not going to bother to specify here.
+                timestamp_info = str(timestamp_ms)
+
+            logger.info(
```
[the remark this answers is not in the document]
timmc-edx: Hmm... do you think it's worth it? It would mean more header-reading code on the producer and consumer sides—not a _lot_ but enough that I would feel the need to add a new try-block around it on the producing side. Something like this:

```
            message_id = next(
                (h[1].decode() for h in msg.headers() if h[0] == HEADER_ID.message_header_key),
                None
            )
```

I guess I'm not actually sure what the role of that field is, since topic+partition+offset is already a unique key we can use to match up and identify messages. Maybe it matters more if we use the outbox pattern, or have DLQs?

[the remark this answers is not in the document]
robrap: Related to our retro conversation, how about one of you write up a ticket for this instead (if you wish), and drop it on to the future milestone? Or, we can add it when and if we need it. It seems we have a start at audit logs either way.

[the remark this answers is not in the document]
robrap: @timmc-edx @rgraber: Thoughts on my suggestion? Wondering how we can close this out. Thanks.

edx_event_bus_kafka/internal/producer.py
```diff
@@ -51,7 +51,7 @@ def record_producing_error(error, context):
         raise Exception(error)
     except BaseException:
         record_exception()
-        logger.exception(f"Error producing event to event bus. {error=!s} {context!r}")
+        logger.exception(f"Error delivering message to Kafka event bus. {error=!s} {context!r}")
```
rgraber, line 54: Make sure to update the docs to reflect this. I'm pretty sure they say to search for "Error producing event"

docs/decisions/0010_audit_logging.rst
timmc-edx: Toggle added.

edx_event_bus_kafka/internal/producer.py
timmc-edx: Good catch, thanks! Added a comment for now.

rgraber (approved):
LGTM

Issue #124, opened 2023-02-01
Add annotations for event bus settings


robrap:
AC:
- [x] Add annotation documentation for the rest of the settings in `config.py`.

Only one of the settings in `config.py` is annotated: https://github.com/openedx/event-bus-kafka/blob/bb4262f64b6361bbc7b1d8c621e7f74270f3bda4/edx_event_bus_kafka/internal/config.py#L88-L95

This ticket is for adding annotations for the rest of the settings in this file.
https://github.com/openedx/event-bus-kafka/blob/main/edx_event_bus_kafka/internal/config.py

We might want to link to the new annotations from:
* https://github.com/openedx/event-bus-kafka#documentation
* https://openedx.atlassian.net/wiki/spaces/AC/pages/3508699151/How+to+start+using+the+Event+Bus#First-time-setup

Note: We could consider updating docs in this library to extract/publish these annotations, as is done in edx-platform. This could be part of this ticket or its own ticket.

robrap:
Should we do this now, or add to Future milestone?

timmc-edx:
Created https://github.com/openedx/event-bus-kafka/issues/139 or followup.

PR #126, opened 2023-02-06
docs: Add docs for all Django settings

timmc-edx:
Addresses https://github.com/openedx/event-bus-kafka/issues/124

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/edx_event_bus_kafka/internal/config.py
+++ b/edx_event_bus_kafka/internal/config.py
@@ -38,12 +38,27 @@ def get_schema_registry_client():
         warnings.warn('Library confluent-kafka not available. Cannot create schema registry client.')
         return None
 
+    # .. setting_name: EVENT_BUS_KAFKA_SCHEMA_REGISTRY_URL
+    # .. setting_default: None
+    # .. setting_description: URL of the Avro schema registry, required for managing the evolution
+    #   of event bus data and ensuring that consumers are able to decode the events that
+    #   are produced. This URL is required for both producers and consumers and must point
+    #   to an instance of Confluent Schema Registry:
+    #   https://docs.confluent.io/platform/current/schema-registry/index.html
     url = getattr(settings, 'EVENT_BUS_KAFKA_SCHEMA_REGISTRY_URL', None)
     if url is None:
         warnings.warn("Cannot configure event-bus-kafka: Missing setting EVENT_BUS_KAFKA_SCHEMA_REGISTRY_URL")
         return None
 
+    # .. setting_name: EVENT_BUS_KAFKA_SCHEMA_REGISTRY_API_KEY
+    # .. setting_default: None
+    # .. setting_description: API key for talking to the Avro schema registry specified in
+    #   ``EVENT_BUS_KAFKA_SCHEMA_REGISTRY_URL``.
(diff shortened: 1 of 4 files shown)


edx_event_bus_kafka/internal/config.py
```diff
@@ -38,12 +38,27 @@ def get_schema_registry_client():
...
+    #   to an instance of Confluent Schema Registry:
+    #   https://docs.confluent.io/platform/current/schema-registry/index.html
     url = getattr(settings, 'EVENT_BUS_KAFKA_SCHEMA_REGISTRY_URL', None)
     if url is None:
         warnings.warn("Cannot configure event-bus-kafka: Missing setting EVENT_BUS_KAFKA_SCHEMA_REGISTRY_URL")
         return None
 
+    # .. setting_name: EVENT_BUS_KAFKA_SCHEMA_REGISTRY_API_KEY
+    # .. setting_default: None
+    # .. setting_description: API key for talking to the Avro schema registry specified in
```
rgraber, line 55: This key and secret are optional as well (figure we should probably be consistent with how we describe the Kafka API key)

rgraber (commented):
One minor but I think worthwhile suggestion

edx_event_bus_kafka/internal/config.py
timmc-edx: Oh! Yes, I was wondering about this and forgot to leave myself a note. The Kafka API key/secret are definitely optional because if either one is omitted, we skip adding both of them to the config. But with these schema registry values, we _always_ use them, even if they're None or empty. So for instance we could end up with a `basic.auth.user.info` of `:`, `user:`, `:pass`, or `user:pass`. Not all of those look particularly sensible. I'll go check the schema registry docs again...

(Also, I see that I've documented them as defaulting to `None` but they're actually `''`. Oops.)

rgraber: This checks if user info = ('','') (ie we passed ':') so I think we're fine https://github.com/confluentinc/confluent-kafka-python/blob/bbe1225dc7d36df2c7e3e0f7476bb283fabcd1ff/src/confluent_kafka/schema_registry/schema_registry_client.py#L101

rgraber: It looks like it will look to the url first for auth information and then error if there is also `basic.auth.user.info.` I vote we punt on what happens if the schema registry server doesn't expect any auth and just put in docs somewhere that user should use the settings provided instead of putting the user and password in the URL

timmc-edx: Yeah, we should probably document that auth goes in the auth settings and not in the URL. But what to do about required/optional? I don't want to make claims in the docstrings that we can't back up, and so I'm tempted to just... not mention required/optional either way, for those two.

rgraber: I think "Optional" still makes sense because we do know for sure that you don't have to always have to set them.

timmc-edx: Hmm. I still don't know what happens if the schema registry isn't configured for auth and we pass an auth header containing `':'` anyway. Maybe nothing? I guess if it causes any issues then we can put in a bugfix when we find out.

rgraber (approved):
LGTM

Issue #135, opened 2023-02-23
KeyError when running event consumer


rgraber:
In both course-discovery and subscriptions (a new service), running the consume_events command with `-s org.openedx.learning.course.enrollment.created.v1` results in `KeyError: 'org.openedx.learning.auth.session.login.completed.v1'`.
Presumably, this is because we have not yet loaded the SESSION_LOGIN_COMPLETED signal, so it has not been added to the OpenEdxPublicSignal cache. What is unclear is why this doesn't cause problems with the COURSE_CATALOG_INFO_CHANGED event.
Regardless, we should probably change the consume_events API to take the loading path of the signal (eg `openedx_events.learning.signals.SESSION_LOGIN_COMPLETED`) instead of the event_type. We can then follow the pattern we use in `produce_event` and load the module to ensure that the signal is present in the OpenEdxPublicSignal mapping.

PR #137, opened 2023-02-24
fix: load all signals in consume_events

rgraber:
Issue: https://github.com/openedx/event-bus-kafka/issues/135
Blocked on https://github.com/openedx/openedx-events/pull/187

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [ ] Commits are squashed

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,13 @@ Change Log
 Unreleased
 **********
 
+[3.9.5] - 2023-02-24
+********************
+
+Fixed
+=====
+* load all signals in consume_events command before running the consumer
+
 [3.9.4] - 2023-02-16
 ********************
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.rst
```diff
@@ -14,6 +14,13 @@ Change Log
 Unreleased
 **********
 
+[3.9.5] - 2023-02-24
+********************
+
+Fixed
+=====
+* load all signals in consume_events command before running the consumer
```
timmc-edx, line 22: It would be nice to explain what the failure mode is that this fixes. Maybe something like this?

```suggestion
* ``consume_events`` command now loads all public signals so that the consumer can load signals by ``event_type`` even if their modules were not already imported
```

PR #157, opened 2023-05-04
Manually deserialize messages from a Consumer instead of using the DeserializingConsumer. This is in service of eventually being able to consume events of multiple types from the same object. Left as Unreleased because it is just the first PR of a series to add the full feature.

rgraber:
**Merge checklist:**
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

### DIFF
--- a/edx_event_bus_kafka/internal/consumer.py
+++ b/edx_event_bus_kafka/internal/consumer.py
@@ -29,9 +29,10 @@ logger = logging.getLogger(__name__)
 # See https://github.com/openedx/event-bus-kafka/blob/main/docs/decisions/0005-optional-import-of-confluent-kafka.rst
 try:
     import confluent_kafka
-    from confluent_kafka import TIMESTAMP_NOT_AVAILABLE, DeserializingConsumer
+    from confluent_kafka import TIMESTAMP_NOT_AVAILABLE, Consumer
     from confluent_kafka.error import KafkaError
     from confluent_kafka.schema_registry.avro import AvroDeserializer
+    from confluent_kafka.serialization import MessageField, SerializationContext
 except ImportError:  # pragma: no cover
     confluent_kafka = None
 
@@ -117,39 +118,27 @@ class KafkaEventConsumer:
         self.consumer = self._create_consumer()
         self._shut_down_loop = False
 
-    # return type (Optional[DeserializingConsumer]) removed from signature to avoid error on import
+    # return type (Optional[Consumer]) removed from signature to avoid error on import
     def _create_consumer(self):
         """
-        Create a DeserializingConsumer for events of the given signal instance.
+        Create a Consumer in the correct consumer group for events of the associated topic.
 
         Returns
             None if confluent_kafka is not available.
             DeserializingConsumer if it is.
         """
 
-        schema_registry_client = get_schema_registry_client()
-
-        signal_deserializer = AvroSignalDeserializer(self.signal)
-
-        def inner_from_dict(event_data_dict, ctx=None):  # pylint: disable=unused-argument
-            return signal_deserializer.from_dict(event_data_dict)
-
         consumer_config = load_common_settings()
 
-        # We do not deserialize the key because we don't need it for anything yet.
-        # Also see https://github.com/openedx/openedx-events/issues/86 for some challenges on determining key schema.
         consumer_config.update({
             'group.id': self.group_id,
-            'value.deserializer': AvroDeserializer(schema_str=signal_deserializer.schema_string(),
-                                                   schema_registry_client=schema_registry_client,
-                                                   from_dict=inner_from_dict),
             # Turn off auto commit. Auto commit will commit offsets for the entire batch of messages received,
             # potentially resulting in data loss if some of those messages are not fully processed. See
             # https://newrelic.com/blog/best-practices/kafka-consumer-config-auto-commit-data-loss
             'enable.auto.commit': False,
         })
 
-        return DeserializingConsumer(consumer_config)
+        return Consumer(consumer_config)
 
     def _shut_down(self):
         """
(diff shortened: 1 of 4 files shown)


edx_event_bus_kafka/internal/consumer.py
```diff
@@ -117,39 +118,27 @@ def __init__(self, topic, group_id, signal):
         self.consumer = self._create_consumer()
         self._shut_down_loop = False
 
-    # return type (Optional[DeserializingConsumer]) removed from signature to avoid error on import
+    # return type (Optional[Consumer]) removed from signature to avoid error on import
```
timmc-edx, line 121: Looks like we should drop the Optional (not a result of this PR, though).

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -381,6 +370,31 @@ def emit_signals_from_message(self, msg):
...
+        """
+        Deserialize an Avro message value
+
+        Arguments:
+            msg (Message): the raw message from the consumer
+
+        Returns:
+            The deserialized message value
+        """
+        schema_registry_client = get_schema_registry_client()
```
timmc-edx, line 383: I believe this can be lifted to `__init__` rather than being created each time.

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -381,6 +370,31 @@ def emit_signals_from_message(self, msg):
...
+        signal_deserializer = AvroSignalDeserializer(self.signal)
+
+        def inner_from_dict(event_data_dict, ctx=None):  # pylint: disable=unused-argument
+            return signal_deserializer.from_dict(event_data_dict)
+
+        # We do not deserialize the key because we don't need it for anything yet.
+        # Also see https://github.com/openedx/openedx-events/issues/86 for some challenges on determining key schema.
+        avro_deserializer = AvroDeserializer(schema_str=signal_deserializer.schema_string(),
+                                             schema_registry_client=schema_registry_client,
+                                             from_dict=inner_from_dict)
```
timmc-edx, line 394: Can we move `avro_deserializer` and the lines above it to a cached function? See `get_serializers` in the producer code, including the `_reset_caches` function at the end of the file. If I recall correctly, creating the deserializer isn't _terribly_ computationally intensive, but far more so than a hash lookup.

edx_event_bus_kafka/internal/tests/test_consumer.py
```diff
@@ -152,6 +155,7 @@ def test_reset_offsets_and_sleep_indefinitely(self):
         EVENT_BUS_KAFKA_SCHEMA_REGISTRY_URL='http://localhost:12345',
         EVENT_BUS_KAFKA_BOOTSTRAP_SERVERS='localhost:54321',
         EVENT_BUS_TOPIC_PREFIX='local',
+        EVENT_BUS_KAFKA_CONSUMER_CONSECUTIVE_ERRORS_LIMIT=4,
```
timmc-edx, line 158: Was this needed, or is this just to match the other tests?

edx_event_bus_kafka/internal/tests/test_consumer.py
```diff
@@ -433,9 +445,12 @@ def test_check_event_error(self):
     @ddt.data(True, False)
     def test_emit_success(self, audit_logging, mock_logger, mock_set_attribute):
         self.signal.disconnect(fake_receiver_raises_error)  # just successes for this one!
+        msg = copy.copy(self.normal_message)
```
timmc-edx, line 448: Is the copy necessary? setUp should be creating a fresh copy for each test.

(I'm fairly sure I was the one who introduced `copy` in the existing unit tests, but now I think it was actually unnecessary...)

rgraber: for the sake of keeping this PR clean I'll keep the legacy ones but fix the new ones

edx_event_bus_kafka/internal/consumer.py
rgraber: accidentally took out the bit that will return none if confluent-kafka isn't available. will fix.

edx_event_bus_kafka/internal/consumer.py
rgraber: Moved it to module initiation

rgraber: Nevermind, moving it to instance initiation because doing it at the top level of the module does weird things to tests. We could just have a test.py environment file but I would put that in a separate PR.

edx_event_bus_kafka/internal/consumer.py
timmc-edx: It looks to me like it already couldn't return None, so I don't know that there's any need to change that.

rgraber: It would make it more consistent with the producer

timmc-edx: I'm not sure we want to be consistent in that direction. It just delays the error until consume/produce time. I suspect the Optional part is a holdover from an earlier design where we didn't have a wrapper around the consumer object. Not sure on that, though.

timmc-edx: (Specifically, we'll get "can't call method poll() on type None" or some other less-useful error.)

edx_event_bus_kafka/internal/consumer.py
```diff
@@ -381,6 +378,20 @@ def emit_signals_from_message(self, msg):
         if AUDIT_LOGGING_ENABLED.is_enabled():
             logger.info('Message from Kafka processed successfully')
 
+    def deserialize_message_value(self, msg):
```
timmc-edx, line 381: Maybe make this private? (prefix with underscore)

edx_event_bus_kafka/internal/consumer.py
rgraber: Updated since I don't think it's worth delaying the PR. We may want to reconsider when we make the consumer more configurable like the producer.

timmc-edx (approved):

PR #162, opened 2023-05-09
Rsgraber/20230508 adr

rgraber:
Add an ADR about how and why we will be allowing multiple event types per topic with the event bus.

Note:
While the ADR obviously posits a decision as made, I am somewhat ambivalent on the proposed solution given the consequences. I would definitely appreciate other opinions, especially as regards either using a SettingsToggle or using Avro unions.

**Merge checklist:**
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

### DIFF
--- a/docs/decisions/0011_multiple_event_types_per_topic.rst
+++ b/docs/decisions/0011_multiple_event_types_per_topic.rst
@@ -72,10 +72,3 @@ Avro unions are a datatype representing the possibility of multiple different sc
 Using Avro unions is currently not feasible because of `a bug`_ in the confluent-kafka-python library. It would also require all schemas, not just the union ones, to be created, evolved, registered independently of event-producing code, requiring separate updates to configurations every time a new event type was added to a topic or we wanted to update an event schema. This is because, in order to use unions, auto-registration of schemas must be disabled, which is done on a per-serializer basis. As mentioned previously, all serializers have the same configurations. Changing this would be a significant lift. 
 
 .. _a bug: https://github.com/confluentinc/confluent-kafka-python/issues/1562
-
-
-
-
-
-
-
--- a/docs/decisions/0011_multiple_event_types_per_topic.rst
+++ b/docs/decisions/0011_multiple_event_types_per_topic.rst
@@ -40,7 +40,7 @@ and the topic ``my-topic``, the serializer will look for a schema registered as 
 Consequences
 ************
 * The record names for each event type will need to be distinct from each other. Originally, they were all just "CloudEvent." We needed to add namespaces to ensure that they were unique. Note that this will be necessary regardless of which solution we choose.
-* Any event type will be able to be written to any topic
+* Any event type will be able to be written to any topic, even for topics where we only want a single event type.
 * Schema evolution rules will be enforced separately for each event type on a topic.
 * The event-bus-kafka consumer will need to be updated to determine the signal from the message headers rather than taking a signal as a passed argument in the management command. This will also be necessary regardless of which solution we choose.
 
--- a/docs/decisions/0011_multiple_event_types_per_topic.rst
+++ b/docs/decisions/0011_multiple_event_types_per_topic.rst
@@ -68,6 +68,7 @@ In theory we could avoid the issue of event types with different schemas on the 
 Avro Unions
 ===========
 Avro unions are a datatype representing the possibility of multiple different schemas for a single field or record. They often contain references to other registered schemas. For example, setting a topic schema to [my.signal.CloudEvent, my.other.signal.CloudEvent] would allow events with either the my.signal.CloudEvent or my.other.signal.CloudEvent schema, but no others. This has the advantage of being configurable by topic and allowing greater control over which events are allowed on a topic.
+
 Using Avro unions is currently not feasible because of `a bug`_ in the confluent-kafka-python library. It would also require all schemas, not just the union ones, to be created, evolved, registered independently of event-producing code, requiring separate updates to configurations every time a new event type was added to a topic or we wanted to update an event schema. This is because, in order to use unions, auto-registration of schemas must be disabled, which is done on a per-serializer basis. As mentioned previously, all serializers have the same configurations. Changing this would be a significant lift. 
 
 .. _a bug: https://github.com/confluentinc/confluent-kafka-python/issues/1562
--- a/docs/decisions/0011_multiple_event_types_per_topic.rst
+++ b/docs/decisions/0011_multiple_event_types_per_topic.rst
@@ -4,15 +4,14 @@
 Status
 ******
 
-- **Accepted** DATE TO BE ADDED BEFORE MERGING
+**Accepted** DATE TO BE ADDED BEFORE MERGING
 
 Context
 *******
 
 Following `this ADR`_ in openedx-events, all implementations of event bus must support multiple event types on the same topic. This ADR explains how we will implement this using Kafka.
-**UPDATE LINK AFTER THAT ADR MERGES**
 
-.. _this ADR: https://github.com/openedx/openedx-events/pull/217
+.. _this ADR: https://openedx-events.readthedocs.io/en/latest/decisions/0010-multiple-event-types-per-topic.html
 
 Decision
 ********
@@ -59,10 +58,6 @@ Usually we would put this sort of feature behind a SettingsToggle so deployers c
 Using different serializers based on the topic
 ==============================================
 We could refactor the code to configure the AvroSerializer differently based on the topic, only changing the subjec name strategy when the topic should allow multiple event types. This would be a significant amount of work and would require some list of which topics accept multiple event types, which would have to be maintained and would reduce the flexibility of the event bus.
-
-Require related event types to have the same schema
-===================================================
-In theory we could avoid the issue of event types with different schemas on the same topic by simply not allowing them, requiring all event types that are intended to go on the same topic to have the same schema. This would require knowing in advance which event types will go to which topics and likely result in lots of extraneous fields which are only necessary for some of the event types and not others.
 
 Avro Unions
 ===========
--- a/docs/decisions/0011_multiple_event_types_per_topic.rst
+++ b/docs/decisions/0011_multiple_event_types_per_topic.rst
@@ -16,7 +16,7 @@ Following `this ADR`_ in openedx-events, all implementations of event bus must s
 
 Decision
 ********
-We will update the code for event serializers, which are responsible for registering and validating schemas against the schema registry, to use the TopicRecordName strategy.
+To comply with the event bus requirements, we will use the TopicRecordName strategy in the event serializers, which are responsible for registering and validating schemas against the schema registry.
 
 TopicRecordName Strategy
 ========================
@@ -24,7 +24,7 @@ The Confluent Schema Registry stores schemas according to one of three strategie
 
 The default strategy the TopicName strategy. Under this strategy, the serializer will determine the existing schema by looking for a schema registered as ``<topic-name>-value`` (``<topic-name>-key`` when serializing the key), the topic name having been passed in the SerializationContext object. Therefore, if the serializer is given a new event type with the same topic, unless the new event type schema happens to be a valid evolution of the existing one, it will fail.
 
-Using the TopicRecordName strategy, schemas are registered by topic and record name. On serializing an event, the serializer looks for a schema registered as ``<topic-name>-<record-name>``.
+Using the TopicRecordName strategy, schemas are registered by topic and record name. The record name is set from the schema as ``<namespace>.<name>`` (or just ``<name>`` if no namespace is provided). On serializing an event, the serializer looks for a schema registered as ``<topic-name>-<record-name>``.
 For example, given a schema like::
 
     {
--- a/docs/decisions/0011_multiple_event_types_per_topic.rst
+++ b/docs/decisions/0011_multiple_event_types_per_topic.rst
@@ -53,9 +53,13 @@ Setting the subject name strategy on the topic rather than the client
 =====================================================================
 This option is only available for locally-hosted or Dedicated Confluent Cloud clusters, and thus not a feasible solution for the whole community.
 
-Using a setting toggle
+Using a SettingsToggle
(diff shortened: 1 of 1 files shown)


docs/decisions/0011_multiple_event_types_per_topic.rst
```diff
@@ -0,0 +1,74 @@
...
+
+    {
+        "doc": "Avro Event Format for CloudEvents created with openedx_events/schema",
+        "fields": [{"name": "str_field", "type": "string"}]
+        "name": "CloudEvent",
+        "namespace": "my.event.type",
+        "type": "record"
+    }
+
+and the topic ``my-topic``, the serializer will look for a schema registered as ``my-topic-my.event.type.CloudEvent`` and use that to determine if the schema being passed is valid. This means that as long as the namespace is different for the schemas for different event types, the serializer will not error when adding new event types. It will still maintain evolution rules within an event type, however, which is desirable behavior.
```
robrap, line 39: It's unclear how exactly `record-name` relates to the `name` field and `namespace` field. Are these both standard fields looked at by something (what?), and is `namespace` used if provided, otherwise `name` is used?

docs/decisions/0011_multiple_event_types_per_topic.rst
```diff
@@ -0,0 +1,74 @@
...
+
+Consequences
+************
+* The record names for each event type will need to be distinct from each other. Originally, they were all just "CloudEvent." We needed to add namespaces to ensure that they were unique. Note that this will be necessary regardless of which solution we choose.
+* Any event type will be able to be written to any topic
```
robrap, line 44: Maybe something like following, to make it clear that this is an unintended consequence?
```suggestion
* Any event type will be able to be written to any topic, even for topics where we only want a single event type.
```

docs/decisions/0011_multiple_event_types_per_topic.rst
```diff
@@ -0,0 +1,74 @@
...
+However, other types of events are not so easily grouped into a single signal. For example, there are different signals for ``XBLOCK_PUBLISHED``, ``XBLOCK_UPDATED``, and ``XBLOCK_DUPLICATED``, and indeed different schemas across the three signals. Routing these signals through different topics could lead to events being processed in a nonsensical order, for example an xblock being duplicated before it's been published. See `Should You Put Several Event Types in the Same Kafka Topic?`_ and `Putting Several Event Types in the Same Topic – Revisited`_ for more information on why it can be useful to group different event types on the same topic.
+
+.. _Should You Put Several Event Types in the Same Kafka Topic?: https://www.confluent.io/blog/put-several-event-types-kafka-topic/
+.. _Putting Several Event Types in the Same Topic – Revisited: https://www.confluent.io/blog/multiple-event-types-in-the-same-kafka-topic/
+
+Decision
+********
+We will update the code for event serializers, which are responsible for registering and validating schemas against the schema registry, to use the TopicRecordName strategy.
```
robrap, line 20: Maybe start with something like:

> We will allow multiple related event types to be produced to a single topic so that they can be ordered appropriately. We will accomplish this by...

docs/decisions/0011_multiple_event_types_per_topic.rst
```diff
@@ -0,0 +1,74 @@
...
+        "type": "record"
+    }
+
+and the topic ``my-topic``, the serializer will look for a schema registered as ``my-topic-my.event.type.CloudEvent`` and use that to determine if the schema being passed is valid. This means that as long as the namespace is different for the schemas for different event types, the serializer will not error when adding new event types. It will still maintain evolution rules within an event type, however, which is desirable behavior.
+
+Consequences
+************
+* The record names for each event type will need to be distinct from each other. Originally, they were all just "CloudEvent." We needed to add namespaces to ensure that they were unique. Note that this will be necessary regardless of which solution we choose.
+* Any event type will be able to be written to any topic
+* Schema evolution rules will be enforced for event types but not for topics.
```
robrap, line 45: I wonder if "but not for topics" would be a little confusing for non-experts? What about something like:
```suggestion
* Schema evolution rules will be enforced separately for each event type on a topic.
```

docs/decisions/0011_multiple_event_types_per_topic.rst
```diff
@@ -0,0 +1,74 @@
...
+
+Using different serializers based on the topic
+==============================================
+We could refactor the code to configure the AvroSerializer differently based on the topic, only changing the subjec name strategy when the topic should allow multiple event types. This would be a significant amount of work and would require some list of which topics accept multiple event types, which would have to be maintained and would reduce the flexibility of the event bus.
+
+Require related event types to have the same schema
+===================================================
+In theory we could avoid the issue of event types with different schemas on the same topic by simply not allowing them, requiring all event types that are intended to go on the same topic to have the same schema. This would require knowing in advance which event types will go to which topics and likely result in lots of extraneous fields which are only necessary for some of the event types and not others.
```
robrap, line 66: Also, this is considered an anti-pattern, so we should definitely avoid this.

docs/decisions/0011_multiple_event_types_per_topic.rst
```diff
@@ -0,0 +1,74 @@
...
+
+and the topic ``my-topic``, the serializer will look for a schema registered as ``my-topic-my.event.type.CloudEvent`` and use that to determine if the schema being passed is valid. This means that as long as the namespace is different for the schemas for different event types, the serializer will not error when adding new event types. It will still maintain evolution rules within an event type, however, which is desirable behavior.
+
+Consequences
+************
+* The record names for each event type will need to be distinct from each other. Originally, they were all just "CloudEvent." We needed to add namespaces to ensure that they were unique. Note that this will be necessary regardless of which solution we choose.
+* Any event type will be able to be written to any topic
+* Schema evolution rules will be enforced for event types but not for topics.
+* The event-bus-kafka consumer will need to be updated to determine the signal from the message headers rather than taking a signal as a passed argument in the management command. This will also be necessary regardless of which solution we choose.
+
```
robrap, line 47: Somewhere could we note that in the future we might be able to detect if a service is only sending one event type for topic for all topics, and to not configure in the same way in this case?

robrap (commented):
I wonder if we need an ADR in openedx-events simply stating that all implementations (including Redis as an example) will need to be able to support sending multiple event types to a single topic to allow for better support of ordering these events. This ADR would then just be about how we implement this other decision in Kafka. Thoughts?

rgraber:
> I wonder if we need an ADR in openedx-events simply stating that all implementations (including Redis as an example) will need to be able to support sending multiple event types to a single topic to allow for better support of ordering these events. This ADR would then just be about how we implement this other decision in Kafka. Thoughts?

I think that's a good idea. I should probably hold off on this ADR so it can point to that one. I'll move some of the text in the context section to the openedx-events ADR.

docs/decisions/0011_multiple_event_types_per_topic.rst
rgraber: [resolved]

docs/decisions/0011_multiple_event_types_per_topic.rst
rgraber: [resolved]

docs/decisions/0011_multiple_event_types_per_topic.rst
rgraber: I'm not sure that's true. See the consequences of using a rollout toggle. We pretty much need this to be all-or-nothing across services.

docs/decisions/0011_multiple_event_types_per_topic.rst
```diff
@@ -0,0 +1,73 @@
+10. Multiple event types per topic
+##################################
+
+Status
+******
+
+- **Accepted** DATE TO BE ADDED BEFORE MERGING
```
robrap, line 7: Nit: You can drop the `-`.

docs/decisions/0011_multiple_event_types_per_topic.rst
```diff
@@ -0,0 +1,73 @@
...
+
+Context
+*******
+
+Following `this ADR`_ in openedx-events, all implementations of event bus must support multiple event types on the same topic. This ADR explains how we will implement this using Kafka.
+**UPDATE LINK AFTER THAT ADR MERGES**
+
```
robrap, line 14: Can you add a small section providing a brief list of possible solutions to this problem in Kafka? Could just be bullets, and you can note that they'll be covered in more detail later. But basically, the context is that there are 2 (or 3, etc. depending on how you look at it) ways of accomplishing this, and this ADR is just about the one we selected.

docs/decisions/0011_multiple_event_types_per_topic.rst
```diff
@@ -0,0 +1,73 @@
...
+.. _this ADR: https://github.com/openedx/openedx-events/pull/217
+
+Decision
+********
+To comply with the event bus requirements, we will use the TopicRecordName strategy in the event serializers, which are responsible for registering and validating schemas against the schema registry.
```
robrap, line 19: Can you summarize why? For example: This method seems to be the simplest solution that solves the problem. (Or whatever you think the summary would be.) You don't really call out the pros/cons of the selected or rejected alternatives, so the why doesn't really jump out at you. Consider adding pros/cons, or reorganizing to make this more clear.

docs/decisions/0011_multiple_event_types_per_topic.rst
robrap: Maybe we can have a chat, simply because I don't follow it. Only a single service should ever produce to a topic. I don't think we have an ADR around that, but that is a pattern that seems to be generally accepted. There might be several instances (e.g. EC2 instances, k8s containers) of the same service, but I don't think that is what you are referring to...

As I was writing this, I had a thought that _may_ be what you were trying to say. The above constraint is possibly about an event type, and not a topic. So, one question is whether this ADR should declare that all event types being produced to a shared topic must be sourced from the same service. I say yes.

Does this have anything to do with the point you were trying to make, or do we also need an entirely separate discussion as well?

docs/decisions/0011_multiple_event_types_per_topic.rst
```diff
@@ -0,0 +1,73 @@
...
+Rejected Alternatives
+*********************
+
+Setting the subject name strategy on the topic rather than the client
+=====================================================================
+This option is only available for locally-hosted or Dedicated Confluent Cloud clusters, and thus not a feasible solution for the whole community.
+
+Using a SettingsToggle
+======================
+Usually we would put this sort of feature behind a SettingsToggle so deployers could enable it and disable it as they chose. However, because the subject name strategy is set at the client rather than topic or cluster level, all clients who produce to the same topic(s) must have the same subject name strategy set. While the consequences of having multiple clients with different subject name strategies writing to the same topic are not 100% clear, at the very least it would result in a proliferation of confusing schemas in the registry.
```
robrap, line 57: I'm not clear yet if this section will stay or not until we discuss on the other comment thread.

**IF** it stays, I wonder if we should:
- Add to the decision that we will **not** provide.a toggle to disable. (may not be necessary)
- Clarify that this is not an alternative to the TopicRecordName Strategy decision, but an implementation detail of that alternative. Maybe this would just move to the decision section altogether, and drop this from here?

docs/decisions/0011_multiple_event_types_per_topic.rst
```diff
@@ -0,0 +1,73 @@
...
+
+Using different serializers based on the topic
+==============================================
+We could refactor the code to configure the AvroSerializer differently based on the topic, only changing the subjec name strategy when the topic should allow multiple event types. This would be a significant amount of work and would require some list of which topics accept multiple event types, which would have to be maintained and would reduce the flexibility of the event bus.
+
+Require related event types to have the same schema
+===================================================
+In theory we could avoid the issue of event types with different schemas on the same topic by simply not allowing them, requiring all event types that are intended to go on the same topic to have the same schema. This would require knowing in advance which event types will go to which topics and likely result in lots of extraneous fields which are only necessary for some of the event types and not others.
```
robrap, line 65: Can you either drop this section, or just point to the other ADR that has the updated language?

robrap (commented):
Thank you. The idea around pros/cons is non-blocking.

docs/decisions/0011_multiple_event_types_per_topic.rst
rgraber: I was thinking about multiple services producing to the same topic. If we are comfortable saying that is not allowed and if you do that you could end up with weird schema problems, that's fine. I would definitely call that out in the ADR though.

docs/decisions/0011_multiple_event_types_per_topic.rst
rgraber: [fixed]

docs/decisions/0011_multiple_event_types_per_topic.rst
rgraber: [fixed] Added back the SettingsToggle, and noted that this requires one producer per topic

docs/decisions/0011_multiple_event_types_per_topic.rst
rgraber: [fixed] removed

docs/decisions/0011_multiple_event_types_per_topic.rst
rgraber: [fixed]

docs/decisions/0011_multiple_event_types_per_topic.rst
rgraber: [fixed]

docs/decisions/0011_multiple_event_types_per_topic.rst
```diff
@@ -0,0 +1,90 @@
+10. Multiple event types per topic
+##################################
+
+Status
+******
+
+**Accepted** DATE TO BE ADDED BEFORE MERGING
```
robrap, line 7: Reminder comment so this isn't forgotten.

robrap (approved):
Thank you!

PR #192, opened 2023-08-25
fix: Reorder context assignment in producer code.

dianakhuang:
In error logging, we did not have the full_topic
available for searching/alerting because of the order in which these assignments happened. Reordering them to try to ensure we always get the full_topic in context.

https://github.com/edx/edx-arch-experiments/issues/411

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/edx_event_bus_kafka/internal/producer.py
+++ b/edx_event_bus_kafka/internal/producer.py
@@ -284,13 +284,17 @@ class KafkaEventProducer(EventBusProducer):
         context = ProducingContext(signal=signal, initial_topic=topic, event_key_field=event_key_field,
                                    event_data=event_data, event_metadata=event_metadata)
         try:
-            context.event_data_as_json = json.dumps(get_signal_serializer(signal).to_dict(event_data))
-            context.event_metadata_as_json = event_metadata.to_json()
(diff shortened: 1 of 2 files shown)


edx_event_bus_kafka/internal/producer.py
```diff
@@ -284,13 +284,17 @@ def send(
         context = ProducingContext(signal=signal, initial_topic=topic, event_key_field=event_key_field,
                                    event_data=event_data, event_metadata=event_metadata)
         try:
-            context.event_data_as_json = json.dumps(get_signal_serializer(signal).to_dict(event_data))
-            context.event_metadata_as_json = event_metadata.to_json()
+            # Please keep these context assignments in order from least-risky to most-risky so that we
```
rgraber, line 287: I don't understand why the order matters, or what we mean by "risky"

rgraber: Can we change "risky" to "likely to error"? I think that would be clearer

dianakhuang: Definitely can change that!

rgraber (approved):
LGTM

PR #193, opened 2023-08-28
feat: Use custom exceptions for recording errors.

dianakhuang:
In the past, we used generic exceptions to record errors to New Relic. This creates custom exceptions for producing and consuming so that we can query New Relic on custom exceptions instead of relying on log messages.

https://github.com/edx/edx-arch-experiments/issues/389

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/edx_event_bus_kafka/internal/consumer.py
+++ b/edx_event_bus_kafka/internal/consumer.py
@@ -85,6 +85,12 @@ class ReceiverError(Exception):
         self.causes = causes  # just used for testing
 
 
+class EventConsumptionException(Exception):
+    """
+    Indicates that we had an issue in event production. Useful for filtering on later.
+    """
+
+
 def _reconnect_to_db_if_needed():
     """
     Reconnects the db connection if needed.
@@ -513,8 +519,8 @@ class KafkaEventConsumer(EventBusConsumer):
         try:
             # This is gross, but our record_exception wrapper doesn't take args at the moment,
             # and will only read the exception from stack context.
-            raise Exception(error)
-        except BaseException:
+            raise EventConsumptionException(error)
+        except EventConsumptionException:
             self._add_message_monitoring(run_context=run_context, message=maybe_kafka_message, error=error)
             record_exception()
             logger.exception(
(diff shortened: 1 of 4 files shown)


edx_event_bus_kafka/internal/consumer.py
```diff
@@ -513,8 +519,8 @@ def record_event_consuming_error(self, run_context, error, maybe_message):
         try:
             # This is gross, but our record_exception wrapper doesn't take args at the moment,
             # and will only read the exception from stack context.
-            raise Exception(error)
-        except BaseException:
+            raise EventConsumptionException(error)
+        except EventConsumptionException:
             self._add_message_monitoring(run_context=run_context, message=maybe_kafka_message, error=error)
             record_exception()
```
timmc-edx, line 525: So will this end up recording the `EventConsumptionException`, or will it record the type of `error`?

timmc-edx: (Well, I guess if we lose the type of `error`, that's _already_ happening, so no complaint here.)

edx_event_bus_kafka/internal/producer.py
```diff
@@ -38,6 +38,10 @@
     confluent_kafka = None
 
 
+class EventProductionException(Exception):
```
timmc-edx, line 41: Should these live in a public module?

edx_event_bus_kafka/internal/consumer.py
dianakhuang: Yeah, we were losing the type of error previously. This allows us to at least see a `EventConsumptionException` recorded as the `error.class` in New Relic.

edx_event_bus_kafka/internal/producer.py
dianakhuang: I don't think it needs to, since this is custom to the producing code, and we already had the pattern of keeping consumer exceptions in the `consumer.py` file. I am open to disagreements, though.

timmc-edx (approved):

PR #201, opened 2023-09-20
feat: Reset request cache before handling event

timmc-edx:
This is a just-in-case measure -- we don't know of any specific code that handles events and relies on the cache. But it's likely to come up at some point.

Ticket: https://github.com/openedx/openedx-events/issues/235

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/edx_event_bus_kafka/internal/consumer.py
+++ b/edx_event_bus_kafka/internal/consumer.py
@@ -10,6 +10,7 @@ from django.conf import settings
 from django.db import connection
 from django.dispatch import receiver
 from django.test.signals import setting_changed
+from edx_django_utils.cache import RequestCache
 from edx_django_utils.monitoring import function_trace, record_exception, set_custom_attribute
 from edx_toggles.toggles import SettingToggle
 from openedx_events.event_bus import EventBusConsumer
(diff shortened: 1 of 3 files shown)


edx_event_bus_kafka/internal/consumer.py
```diff
@@ -278,8 +304,10 @@ def _consume_indefinitely(self):
                     msg = self.consumer.poll(timeout=CONSUMER_POLL_TIMEOUT)
                     if msg is not None:
                         with function_trace('_consume_indefinitely_consume_single_message'):
-                            # Before processing, make sure our db connection is still active
-                            _reconnect_to_db_if_needed()
+                            # Before processing, make sure our application state is similar to
+                            # that of a new Django request. (Mimic setup/teardown.)
```
robrap, line 308: Non-blocking question: By "setup/teardown", are you referring to something related to Django requests, or just unit tests or some such? If Django requests, I have https://github.com/openedx/openedx-events/issues/236 open, because I'm wondering if there isn't a more general solution to this problem.

robrap (approved):

edx_event_bus_kafka/internal/consumer.py
robrap: 1. It wouldn't hurt to link to that issue and add some clarity.
2. It actually feels like `prepare_for_new_work_cycle` probably belongs in openedx-events, which would give this benefit to Redis for free. Maybe we could create an issue for that, and let the Redis team know about it. In place of copying this change, maybe they can implement a more shareable solution?

timmc-edx: Created https://github.com/openedx/openedx-events/issues/270

PR #260, opened 2024-07-31
fix: Add monitoring information inside the function trace.

dianakhuang:
Moved function_trace call to a higher level to capture more information about the overall inner loop of the consumer. Also renamed the function trace to ``consumer.consume`` so it's clearer from the outside what the span is doing.

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 **********
 
+[5.8.0] - 2024-08-01
+********************
+Changed
+=======
+* Monitoring: Ensure that we have a root span for each iteration of the consume loop; renamed the trace name to be ``consumer.consume``.
+
 [5.7.0] - 2024-03-22
 ********************
 Changed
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 **********
 
+[5.7.1] - 2024-07-31
```
timmc-edx, line 17: Probably worth a minor version bump (here and in init file).

```suggestion
[5.8.0] - 2024-08-01
```

CHANGELOG.rst
```diff
@@ -14,6 +14,12 @@ Change Log
...
+********************
+Changed
+=======
+* Refactor monitoring to the consumer loop and renamed the trace name to be ``consumer.consume``.
```
timmc-edx, line 21: We could make this more explicit:

```suggestion
* Monitoring: Ensure that we have a root span for each iteration of the consume loop; renamed the trace name to be ``consumer.consume``.
```

timmc-edx (commented):
Some suggestions, nothing blocking.

timmc-edx (approved):

